PR #114, opened 2014-12-23
Snippet Injection for Function Autocompletion

JohannWeging:
Functions are completet as snippets with its parameters as tabstops.
Exampels:
  func foo(bar string) -> foo(${1:bar string})
Fuctions as parameters are allowed too:
  func foo(f func(bar int), baz int) ->
    foo(${1:f func(${2:bar int})}, ${3:baz int})

### DIFF
--- a/lib/gocodeprovider.coffee
+++ b/lib/gocodeprovider.coffee
@@ -1,3 +1,4 @@
+{Range}  = require('atom')
 _ = require('underscore-plus')
 path = require('path')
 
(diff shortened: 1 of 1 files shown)


joefitzgerald:
Thank you for your contribution! This inspired me during the rewrite of `autocomplete-plus`, such that the ability to insert snippets will be handled by `autocomplete-plus` directly. Thanks so much for thinking out of the box on this.

JohannWeging:
I rebased my changes and cleaned up the commit history.
The function completion now uses the build in snippet magic of autocomplete-plus

PR #131, opened 2015-01-14
Docker ci

kn1kn1:
I added Dockerfile and some workarounds in the specs because mainModule variable become null. 
Many failures are now disappeared but there still are 3 failures.
See the log when I build with docker:  https://gist.github.com/kn1kn1/d555835575dfce5cf76b 
The `advanceClock` function might have different behavior between desktop and headless linux, but I do not know the details...


joefitzgerald:
Thanks for the contribution. I'm interested in having a working Dockerfile in the repo, but haven't been able to get this to work. Happy to reopen if you now have a working example.

PR #132, opened 2015-01-16
Fix Depcrecation Warnings

BrennanAtWork:
Just noticed the deprecation cop called some errors out.  Pretty straightforward fixes.  Thanks!

### DIFF
--- a/keymaps/go-plus.cson
+++ b/keymaps/go-plus.cson
@@ -1,5 +1,5 @@
 # For more detailed documentation see
 # https://atom.io/docs/latest/advanced/keymaps
-'.editor':
+'.atom-text-editor':
   'ctrl-alt-c': 'golang:gocover'
   'ctrl-alt-shift-c': 'golang:cleargocover'
--- a/keymaps/go-plus.cson
+++ b/keymaps/go-plus.cson
@@ -1,5 +1,5 @@
 # For more detailed documentation see
 # https://atom.io/docs/latest/advanced/keymaps
-'.atom-text-editor':
+'atom-text-editor':
   'ctrl-alt-c': 'golang:gocover'
   'ctrl-alt-shift-c': 'golang:cleargocover'
--- a/stylesheets/go-plus.less
+++ b/stylesheets/go-plus.less
@@ -7,19 +7,19 @@
 .go-plus {
 }
 
-.editor .gutter .go-plus-error {
+.atom-text-editor::shadow .gutter .go-plus-error {
   color: @text-color-error;
 }
 
-.editor .gutter .git-line-added.go-plus-error {
+.atom-text-editor::shadow .gutter .git-line-added.go-plus-error {
   color: @text-color-error;
 }
 
-.editor .gutter .git-line-modified.go-plus-error {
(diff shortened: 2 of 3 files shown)


BrennanAtWork:
Looks like App Veyor gives an error:

```
apm install autocomplete-plus
'apm' is not recognized as an internal or external command,
operable program or batch file.
```

I'm not sure how this could be related to my changes, does this just happen sometimes?

stylesheets/go-plus.less
```diff
@@ -7,19 +7,19 @@
 .go-plus {
 }
 
-.editor .gutter .go-plus-error {
+.atom-text-editor::shadow .gutter .go-plus-error {
```
stijnvn: It should be `atom-text-editor::shadow` (without the dot). See [Upgrading Your UI Theme](https://atom.io/docs/latest/upgrading/upgrading-your-ui-theme).

stijnvn:
The other css tags should also be prefixed with `atom-text-editor::shadow` to get coverage highlighting working. Or maybe even better, we should just rename the file to `go-plus.atom-text-editor.less`. Then we don't need to change the contents of the file. It will be loaded in the shadow DOM context of the text editor (see [Context-Targeted Style Sheets](https://atom.io/docs/latest/upgrading/upgrading-your-ui-theme)). 

Maybe somebody with more knowledge about this can confirm this, but at least for me, these changes seems to fix the deprecation warnings and the missing highlights.

keymaps/go-plus.cson
```diff
@@ -1,5 +1,5 @@
 # For more detailed documentation see
 # https://atom.io/docs/latest/advanced/keymaps
-'.editor':
+'.atom-text-editor':
```
stijnvn: It should be `atom-text-editor' without the dot. See my other comment.

BrennanAtWork:
@stijnvn,  Thanks for the comments.  According to the upgrade guide you've linked, it says 
> During the transition phase, style sheets targeting the atom-text-editor context will also be loaded globally. Make sure you update your selectors in a way that maintains compatibility with the shadow DOM being disabled.

So it looks like go-plus should support both methods during the transition phase.  This could be done by clone the existing .less file into  `go-plus.atom-text-editor.less`, which doesn't need to change.  Then by making the changes in the existing file to target the shadow DOM using `::shadow`.  This way, when the transition period ends, this file can be simply deleted.

Thoughts?

joefitzgerald:
@BrennanAtWork my advice is to not maintain compatibility with the shadow dom being disabled. It's enabled by default now, and consequently it's ok if someone doesn't see gutter markers if they do disable it.

BrennanAtWork:
Thanks @joefitzgerald, I've simply renamed the `.less` file to target the shadow DOM and changed the `.cson` file as well.  The App Veyor still seems to be failing when it looks for `apm`.  Have you seen this before?

joefitzgerald:
Yeah the windows build has been broken for some time. It's due to Chocolatey / go weirdness and I haven't had time to debug it.

BrennanAtWork:
Ah, I understand.  Well let me know if there's anything else I should change for this PR.

Issue #126, opened 2015-01-13
Run the Go format tool on unsaved buffers.


stijnvn:
It seems that the format tool only works on saved buffers.
Running gofmt on a not properly formatted file does nothing.

joefitzgerald:
`gofmt` does not function on an invalid .go file. However, it should be possible to run gofmt on an unsaved buffer via the menu or a keybinding - that's a valid enhancement.

joefitzgerald:
This is possible as of 4.x. Please update to `go-plus` `v4.1.0`.

PR #133, opened 2015-01-17
Run golang:gofmt on an unsaved buffer.

stijnvn:
I did an attempt to implement the `gofmt` on unsaved buffers as reported in issue #126. 

The unsaved buffer is written to a temporary file. Then `gofmt` is run on that temporary file and its content is read again into the current buffer.

I haven't done any CoffeeScript nor Atom package development before. So feedback is welcome.

### DIFF
--- a/lib/gofmt.coffee
+++ b/lib/gofmt.coffee
@@ -2,6 +2,7 @@
 {Subscriber, Emitter} = require 'emissary'
 _ = require 'underscore-plus'
 path = require 'path'
+fs = require 'fs'
 
 module.exports =
 class Gofmt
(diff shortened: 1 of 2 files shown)


joefitzgerald:
I'm wondering if it would be easier to just have the command force a save of the file if the buffer is modified. In fact, it definitely would be easier, more consistent, and it would also ensure that there is no weirdness because the temporary file is not in $GOPATH.

stijnvn:
I'm not certain it is always desirable to force a save of the file. The use case I had in mind was the following: While I'm coding, I should be able to format the new parts of the code I have written, but at the end, maybe I change my mind and will go back to the previous file. By closing the file and not saving it, I am able to do that.

But you are probably right it is more consistent to just force a save of the file.

joefitzgerald:
> While I'm coding, I should be able to format the new parts of the code I have written, but at the end, maybe I change my mind and will go back to the previous file. 

That's why we use version control software like git :)

joefitzgerald:
Also, atom's undo function should facilitate going back in time also.

stijnvn:
Ok, I will close this pull request then.

Still I find it confusing that running `golang:Gofmt` does nothing on an unsaved buffer. Would you accept a pull request which forces saving the buffer?

joefitzgerald:
Yes! The reason the command is there is for people who have turned off format on save (who really does that, though?). It makes complete sense to detect an unsaved buffer and save it prior to running the format tool. We just need to be wary of autosave being enabled.

By the way, thank you for your contribution(s)! Please don't be discouraged by my comments - I'm very grateful for your involvement in advancing the package's functionality!!

PR #134, opened 2015-01-18
Automatically save the current buffer when running go command

stijnvn:
Automatically save the current buffer when running go build, cover, fmt, lint or vet.

### DIFF
--- a/lib/gobuild.coffee
+++ b/lib/gobuild.coffee
@@ -45,6 +45,9 @@ class Gobuild
       @emit @name + '-complete', editor, saving
       callback(null)
       return
+    unless saving
+      @dispatch.dispatching = true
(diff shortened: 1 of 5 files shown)


stijnvn:
The `@dispatch.dispatching = true` line seems a bit dirty to me. Also the `unless saving` snippet is duplicated 5 times. Maybe I should extend the executor interface with a `needsSaving` parameter or set some env variable and move the saving logic to the executor.
@joefitzgerald what do you think?

joefitzgerald:
Given the extensive modifications that have been made to go-plus in the time since this PR was submitted, I believe this is no longer needed. Thank you so much @stijnvn for your contribution, and sorry that it never was merged.

Issue #135, opened 2015-01-22
Deprecated selectors


alex:
(The only way I could think to show this was a screenshot, sorry!)

![screen shot 2015-01-22 at 3 11 46 pm](https://cloud.githubusercontent.com/assets/772/5863637/01c20c54-a249-11e4-845e-ff8d7753e217.png)

joefitzgerald:
See: #132.

joefitzgerald:
Thanks for reporting, there's a PR with a fix for this that I need to review and merge.

PR #139, opened 2015-01-26
Update deprecated tags and use shadow DOM

lBilali:
1. ".editor" has changed  to "atom-text-editor"
2. Add support for Shadow DOM
as recomended here: https://atom.io/docs/v0.175.0/upgrading/upgrading-your-ui-theme


joefitzgerald:
Thanks for the contribution! This is a duplicate of #135, which has been merged but not released.

Issue #125, opened 2015-01-09
Saving an empty .go file, an error occurred


kyoh86:
Uncaught TypeError: Cannot read property '1' of null
/Users/kyoh/.atom/packages/go-plus/lib/gobuild.coffee:68
Hide Stack Trace
TypeError: Cannot read property '1' of null
  at Gobuild.module.exports.Gobuild.checkBuffer (/Users/kyoh/.atom/packages/go-plus/lib/gobuild.coffee:68:27)
  at /Users/kyoh/.atom/packages/go-plus/lib/dispatch.coffee:299:20
  at /Applications/Atom.app/Contents/Resources/app/node_modules/less-cache/node_modules/less/node_modules/request/node_modules/form-data/node_modules/async/lib/async.js:570:21
  at /Applications/Atom.app/Contents/Resources/app/node_modules/less-cache/node_modules/less/node_modules/request/node_modules/form-data/node_modules/async/lib/async.js:249:17
  at /Applications/Atom.app/Contents/Resources/app/node_modules/less-cache/node_modules/less/node_modules/request/node_modules/form-data/node_modules/async/lib/async.js:125:13
  at Array.forEach (native)
  at _each (/Applications/Atom.app/Contents/Resources/app/node_modules/less-cache/node_modules/less/node_modules/request/node_modules/form-data/node_modules/async/lib/async.js:46:24)
  at async.each (/Applications/Atom.app/Contents/Resources/app/node_modules/less-cache/node_modules/less/node_modules/request/node_modules/form-data/node_modules/async/lib/async.js:124:9)
  at _asyncMap (/Applications/Atom.app/Contents/Resources/app/node_modules/less-cache/node_modules/less/node_modules/request/node_modules/form-data/node_modules/async/lib/async.js:248:13)
  at Object.map (/Applications/Atom.app/Contents/Resources/app/node_modules/less-cache/node_modules/less/node_modules/request/node_modules/form-data/node_modules/async/lib/async.js:219:23)
  at _parallel (/Applications/Atom.app/Contents/Resources/app/node_modules/less-cache/node_modules/less/node_modules/request/node_modules/form-data/node_modules/async/lib/async.js:568:20)
  at Object.async.parallel (/Applications/Atom.app/Contents/Resources/app/node_modules/less-cache/node_modules/less/node_modules/request/node_modules/form-data/node_modules/async/lib/async.js:598:9)
[...]

PR #146, opened 2015-02-02
Fix for error when saving test files / empty go files

fourcube:
Proposed fix for joefitzgerald/go-plus#125

### DIFF
--- a/lib/gobuild.coffee
+++ b/lib/gobuild.coffee
@@ -66,7 +66,7 @@ class Gobuild
     if buffer.getPath().match(/_test.go$/i)
       pre = /^\w*package ([\d\w]+){1}\w*$/img # Need To Support Unicode Letters Also
       match = pre.exec(buffer.getText())
-      testPackage = match[1]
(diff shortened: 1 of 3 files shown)


fourcube:
@joefitzgerald this also fixes #153.

joefitzgerald:
LGTM :+1:

joefitzgerald:
Actually, why are we running gofmt synchronously?

fourcube:
When gofmt reformats the file with the "-w" flag and the execution happens asynchronously, I noticed that the following things happen: 

- The exec callback is invoked
- The buffer is empty (split second)
- All other tools start running ([this spot]) and `gobuild` throws an error (#153, #125 )
- The buffer displays the formatted content

Atom seems to update the buffer's content twice when running `gofmt` asynchronously, whereas the buffer is only updated once, when `gofmt` is executed synchronously.

[this spot]: https://github.com/fourcube/go-plus/blob/533bec5bb6404420ce88332075823e82941058fb/lib/dispatch.coffee#L286

PR #168, opened 2015-03-10
Emit Snippets In Autocomplete

joefitzgerald:
Emits snippets for autocomplete suggestions that are functions. Currently, this is only for function arguments, but could be expanded in the future to include result parameters / types.

This also upgrades gocodeprovider to be compatible with v2.0.0 of the autocomplete+ provider API.

* Fixes #114


joefitzgerald:
Atom's Windows installer is broken. Going to merge this with a broken Windows build.

PR #178, opened 2015-03-26
Ignore more types of whitespace

### LINKED ISSUE #177: Autocomplete triggering on comma
You can't really see my cursor in this screenshot but it's right after I type the comma that the autocomplete window pops up. Annoying mainly because I so often want to hit enter to go to the next line right after typing a comma (instead I get an arbitrary autocomplete).

screen shot 2015-03-26 at 3 25 42 pm

philk:
autocomplete-plus was activating on `,`, `:`, and `}`. I couldn't come up with
a reason why anyone would want to complete on those characters so I
added them to the list. Forgive me if this is bad coffeescript...

Fixes #177

### DIFF
--- a/lib/go-plus.coffee
+++ b/lib/go-plus.coffee
@@ -102,6 +102,12 @@ module.exports =
       type: 'boolean'
       default: true
       order: 16
+    suppressAutocompleteCharacters:
+      title: 'Suppress Autocomplete After Specified Characters'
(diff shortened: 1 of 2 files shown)


joefitzgerald:
Thanks! I enhanced the way configuration is done, and this is released in `v3.3.2`.

PR #180, opened 2015-04-08
Fixing and extending the autocomplete suppression

svanharmelen:
Without the `else char` in `lib/gocodeprovider.coffee` only `space` and `comma` will be added to `@suppressForCharacters`.

Next to that also added a few logical (IMO) defaults and added a way to be able to specify a newline as well.

### DIFF
--- a/lib/go-plus.coffee
+++ b/lib/go-plus.coffee
@@ -105,9 +105,9 @@ module.exports =
       order: 17
     suppressAutocompleteActivationForCharacters:
       title: 'Suppress Autocomplete Activation For Characters'
-      description: 'Suggestions will not be provided when you type one of these characters.'
+      description: 'Suggestions will not be provided when you type one of these characters. Expects a comma separated list with any of the following labels \'comma\', \'space\', \'newline\' or \'tab\' and a string of any additional characters to suppress.'
       type: 'array'
-      default: [')', 'space', ';', 'comma', ':']
+      default: ['comma', 'newline', 'space', 'tab', '/\\()"\':;<>~!@#$%^&*|+=[]{}`?-']
       items:
         type: 'string'
       order: 18
--- a/lib/gocodeprovider.coffee
+++ b/lib/gocodeprovider.coffee
@@ -16,9 +16,13 @@ class GocodeProvider
       @suppressForCharacters = _.map value, (c) ->
         char = c?.trim() or ''
         char = switch
-          when char.toLowerCase() is 'space' then ' '
           when char.toLowerCase() is 'comma' then ','
+          when char.toLowerCase() is 'newline' then '\n'
+          when char.toLowerCase() is 'space' then ' '
+          when char.toLowerCase() is 'tab' then '\t'
+          else char.split ''
         return char
+      @suppressForCharacters = _.flatten(@suppressForCharacters)
       @suppressForCharacters = _.compact(@suppressForCharacters)
 
   setDispatch: (dispatch) ->


joefitzgerald:
LGTM.

svanharmelen:
@joefitzgerald cool... So this will be in the next release then? Already fixed it locally of course, but radar use the mainstream releases...

zmb3:
Might as well add all operators to the defaults, right?

We're currently missing: `+`, `-`, `*`, `%`, `&`, `|`, `^`, `<`, `>`

svanharmelen:
@zmb3 yes indeed... and all quotes (single, double, backtick) should also be added...

I was actually planning on adding a boolean option to just include all configured `Non Word Characters` as that would maybe be a cleaner solution. But I'm wondering if that should go in this package or in the `autocomplete-plus` package...

joefitzgerald:
> But I'm wondering if that should go in this package or in the autocomplete-plus package...

This package. `autocomplete-plus` is used in novel use cases you or I cannot foresee - it shouldn't suppress a suggestion request. Providers are responsible for deciding when to "pass" on a suggestion request.

svanharmelen:
Check... So then I will proceed with adding the feature in here.

Hope to have some time later today or tomorrow.

svanharmelen:
@joefitzgerald just had some time to update this PR and think it looks good now. It makes the autocomplete suppression easier and cleaner to config and use.

Would be great if you can merge this one as it fixes the current implementation which fails to suppress any chars other than ' ' and ','

joefitzgerald:
@svenharmelen thanks. Can you remove the additional preference and move back to the approach of using "tab", "space", etc. for suppression of white space. 

There may be scenarios where someone wants to trigger on some white space but not other white space. E.g. I do want to trigger on space, but not on return or tab.

svanharmelen:
@joefitzgerald ok... Would it then be ok to make one input box with a comma separated list of white space names you want to suppress and one like in the current PR with a string of chars you want to suppress?

Having one box makes it a little dirty IMO and it will generated a pretty long string if you want to have all chars and white spaces in there. Additionally you don't have to treat the comma as a special case. So would that approach be OK?

joefitzgerald:
You should use an array, the way it was before. That requires treating the comma as a special case, and space. Yes, I want on preference fort his, and I want the default to cover most users' desired behavior. So the reality is that few people will interact with it anyway.

joefitzgerald:
Also please rebase on master - it'll fix the test failures on TravisCI.

svanharmelen:
@joefitzgerald I fail to see why having two input boxes would be a bad thing as it only makes things more clean IMO...

Just refactored it according to your comments, but now the defaults is to large for the input box. So users will not be able to see the defaults correctly and will need to look into the code to actually see what chars are there by default:

image

There are about 10 more chars which you now cannot see. Still sure this is how you want to it be? If so I'll rebase and commit this version.

joefitzgerald:
A display issue in settings-view can be fixed by making the input box multi-line. Let's open an issue over there to fix the display issue.

svanharmelen:
@joefitzgerald long story, but the commits are squashed back into one commit so I guess this one should be good to go...

svanharmelen:
@joefitzgerald yet another approach... IMO this is the best solution we've seen so far and it's a mix of both your comments and mine. So really hope you are OK with this one.

Next to it just looking a lot cleaner and more usable to me, it also reduces the need to alter the input box to a multiline box. For comparison this is how the input looks like now:

image

joefitzgerald:
@svenharmelen I think you might be fixating on the length of the field a bit too much. Let's let atom's internals do what they are supposed to do with an array, and not add extra magic to this. You need to also remember some people work with their config file directly.

I made a couple of changes to your suggestion from yesterday, and am just trying to fix a flakey spec prior to merging.

svanharmelen:
@joefitzgerald no it's not the length I'm fixated on. It's on usability and aesthetics. I don't know, maybe it's the mixing of labels with chars but somehow it just doesn't look and feel right to me...

Don't really see how this would be a problem for people working with their config file directly? Would be just as clean and simple for them too...

But it's clear that in your opinion there is only one solution for this, so I will stop trying to offer alternatives. Shall I revert the commit to the state I ended up with yesterday? Or do I understand correctly that you didn't plan on merging that one as well and are working on an slightly adjusted version yourself?

svanharmelen:
@joefitzgerald ah, you already pulled it in... Looks good, thx...

PR #202, opened 2015-05-25
Remove step to install autocomplete-plus

manucheri:
autocomplete-plus is bundled with Atom since 0.199.0, so this step is no longer necessary.


joefitzgerald:
Thanks!

PR #203, opened 2015-05-29
Automatically find Godeps and add it to GOPATH (resolve #53)

### LINKED ISSUE #53: Integrate With Godep (https://github.com/tools/godep)
This was mentioned by [mrwong here](https://github.com/joefitzgerald/go-plus/issues/40#issuecomment-51745957) earlier today, but it might not get noticed in a closed issue.

The request is to have a file (like `.goplus`) that defines what the GOPATH (and other env variables) are *per-project*. While having multiple go paths is not necessarily idiomatic, many people do use this approach. Having the option to change to a different GOPATH based on the project would be incredibly helpful.

rengawm:
When using Godeps, the GOPATH needs to include <project dir>/Godeps/_workspace to access any dependencies provided by Godeps.

This change should resolve https://github.com/joefitzgerald/go-plus/issues/53

### DIFF
--- a/lib/go.coffee
+++ b/lib/go.coffee
@@ -48,17 +48,22 @@ class Go
     # Whenever we build GOPATH, go up the directory tree to identify where there
     # might be a Godeps directory. If we find it in any parent/ancestor directory,
     # add it to GOPATH.
-    goDepCheck = ""
     goDepDir = null
     if cwd?
       goDepDir = @goDepLookupCache.get(cwd)
       if goDepDir is undefined
-        cwdParts = cwd.split(path.sep)
-        for cwdPart in cwdParts
-          if cwdPart.length > 0
-            goDepCheck = goDepCheck + path.sep + cwdPart
-            if fs.existsSync(goDepCheck + path.sep + "Godeps")
-              goDepDir = goDepCheck + path.sep + "Godeps" + path.sep + "_workspace"
+        minLastSeparator = 0
+        if os.platform() is 'win32' and cwd.contains(':\\')
+          minLastSeparator = 2
+
+        goDepCheck = cwd
+        lastSeparator = cwd.lastIndexOf(path.sep)
+        while lastSeparator > minLastSeparator
+          goDepCheck = cwd.substring(0, lastSeparator)
+          if fs.existsSync(goDepCheck + path.sep + "Godeps")
+            goDepDir = goDepCheck + path.sep + "Godeps" + path.sep + "_workspace"
+            break
+          lastSeparator = goDepCheck.lastIndexOf(path.sep)
         @goDepLookupCache.set(cwd, goDepDir)
 
     if goDepDir?
--- a/lib/go.coffee
+++ b/lib/go.coffee
@@ -33,7 +33,7 @@ class Go
     return false unless fs.existsSync(@executable)
     return fs.realpathSync(@executable)
 
-  buildgopath: ->
+  buildgopath: (cwd) ->
     result = ''
     gopathConfig = atom.config.get('go-plus.goPath')
     environmentOverridesConfig = atom.config.get('go-plus.environmentOverridesConfiguration')? and atom.config.get('go-plus.environmentOverridesConfiguration')
@@ -42,6 +42,22 @@ class Go
     result = gopathConfig if not environmentOverridesConfig and gopathConfig? and gopathConfig.trim() isnt ''
     result = gopathConfig if result is '' and gopathConfig? and gopathConfig.trim() isnt ''
     result = result.replace('\n', '').replace('\r', '')
+
+    # Whenever we build GOPATH, go up the directory tree to identify where there
+    # might be a Godeps directory. If we find it in any parent/ancestor directory,
+    # add it to GOPATH.
+    goDepCheck = ""
+    goDepDir = null
+    if cwd?
+      cwdParts = cwd.split(path.sep)
+      for cwdPart in cwdParts
+        if cwdPart.length > 0
+          goDepCheck = goDepCheck+path.sep+cwdPart
+          if fs.existsSync(goDepCheck+path.sep+"Godeps")
+            goDepDir = goDepCheck+path.sep+"Godeps"+path.sep+"_workspace"
+    if goDepDir?
+      result = goDepDir+path.delimiter+result
+
     return @pathexpander.expand(result, '')
 
   splitgopath: ->
--- a/lib/go.coffee
+++ b/lib/go.coffee
@@ -52,9 +52,9 @@ class Go
       cwdParts = cwd.split(path.sep)
       for cwdPart in cwdParts
         if cwdPart.length > 0
-          goDepCheck = goDepCheck+path.sep+cwdPart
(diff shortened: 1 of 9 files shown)


joefitzgerald:
This is great! The only area of concern I have is that (depending on the directory depth) this could be a relatively expensive operation that is on the critical path for typing (because of gocodeprovider, which could be called on every keystroke).

Perhaps we could look to cache the gopath? At least in gocodeprovider?

rengawm:
Yep, fair enough - now they should cache for 2 - 2.5 minutes.

rengawm:
@joefitzgerald Not sure what's up with the CI builds... the test I added is passing, but one is failing which looks like it's unrelated to this change (if I missed something though, let me know). And the build timed out on the Windows box.

rhcarvalho:
What happens when you open two distinct projects that use Godep? AFAICT the GOPATH is set globally for the Atom instance, but in that case we probably better have a per buffer GOPATH? I think that's what vim-go does.

rengawm:
Yeah, this will get the current buffer's file path and look in its directory for a Godeps folder. If it doesn't find one, it looks up one level, then up two levels, and so on. If it finds one, it adds it to the global Atom gopath just for whatever command it's invoking.

rhcarvalho:
I think this will mess up the GOPATH. Image you work on two projects, A and B, and A depends on B (B is vendored under A/Godeps/_workspace).
Now when you are editing a file from project A, you want go-plus tools to use the vendored version of B, while when working on a file from project B you want the tools to use the non-vendored project from your original GOPATH.
Makes sense?

rengawm:
When you're editing a file inside Project A, the gopath will be  ```<path to ProjectA>/Godeps/_workspace:<Atom's global GOPATH>```

When you're editing a file inside Project B, the gopath will be  ```<path to ProjectB>/Godeps/_workspace:<Atom's global GOPATH>``` 

So, when editing code in Project B, it will never use anything from Project A's Godeps, and vice-versa.

I think there is a bit of subjective behavior here, though: Since the global GOPATH is put after the Godeps directory in the final GOPATH, Go will prefer code in the global GOPATH to code in my Godeps dir. I did this intentionally so that you can make changes in Project B and have them immediately available for autocomplete, on-save compile checking, etc, without having to commit them and update your Godeps to include a newer revision. This differs from Godeps' normal behavior when invoked on the command line (as in "godep go build ..."), which is to put Godeps after the existing GOPATH and therefore prefer packages in the Godeps folder to any similar packages in the existing GOPATH.

It's easy enough to change this commit so that it behaves exactly as Godeps does (preferring anything inside Godeps to anything in the rest of the workspace), but I figured the current behavior would be slightly more intuitive. At least for me, when I'm making changes to multiple projects with interdependencies, I generally like to be able to test all of the changes together before committing any of them. I could copy my changes into all the necessary Godeps to do this, but that would be pretty annoying.

rhcarvalho:
@rengawm you're right, I was reading this on mobile and haven't noticed that the cache of gopath was per working dir.

Regarding the order of paths in the final GOPATH, AFAIK the tools will consider paths in GOPATH from left to right, so that `<path to ProjectA>/Godeps/_workspace:<Atom's global GOPATH>` would make the Godeps the preferred code for completion etc.

For the sort of projects I work on, I wouldn't like changes to the upstream project B to apply immediately to project A. I'd prefer to manually update the version of B vendored in A. Afterall, keeping a stable version of the dependencies is one of the reasons for vendoring.

rengawm:
You've convinced me that letting Godeps take priority makes sense in the vendoring world, which I'm less accustomed to so far, heh. But especially given that Godeps works that way when invoked from command-line, it is probably better to be consistent with that behavior.

I did some quick testing using this change as-is, and it did seem to be using the code in my workspace over the code in Godeps at the time, but now I can't reproduce that behavior. (I guess I shouldn't have tested it before coffee, heh.) I couldn't find much clarity on order of precedence when multiple directories are listed in GOPATH in official docs, but I'd agree with your expectation that ```Path1:Path2``` would mean Path1 takes precedence.

So, that back-and-forth... I think we're good here?

rhcarvalho:
Yeah, the current order in GOPATH looks fine. The [docs](http://docs.golang.org/pkg/cmd/go/#hdr-GOPATH_environment_variable) says:

> Go searches each directory listed in GOPATH to find source code, but new packages are always downloaded into the first directory in the list.

Maybe not explicit but "searches each directory listed" implies it does do in the "natural" order, left-to-right. What is "natural" might be culturally arguable, but that behavior is how PATH works on UNIX.

Your PR looks pretty good, I'm looking forward to it being merged :smile:

lib/go.coffee
```diff
@@ -42,6 +45,55 @@ class Go
...
+            goDepDir = goDepCheck + path.sep + "Godeps" + path.sep + "_workspace"
+            break
+          lastSeparator = goDepCheck.lastIndexOf(path.sep)
+        @goDepLookupCache.set(cwd, goDepDir)
+
+    if goDepDir?
+      result = goDepDir + path.delimiter + result
+
+    # Whenever we build GOPATH, go up the directory tree to identify where there
+    # might be a .gopack directory. If we find it in any parent/ancestor directory,
```
rhcarvalho: What tool is using the `.gopack` directory? I couldn't find it. https://github.com/d2fn/gopack seems to use a `gopack.config` file...

rengawm: gopack has you list the dependencies in gopack.config, but it then downloads them to a .gopack directory which you can commit as vendored dependencies.

mfojtik:
+1 for this feature

karalabe:
+1

darinkrauss:
+1

stevenmiller888:
+1

bryanl:
@rengawm and @joefitzgerald is this feature dead?

rengawm:
@bryanl I've been using it in Atom with dev mode, periodically rebasing it onto master... I'll bounce that question to @joefitzgerald :)

joefitzgerald:
Definitely not dead. I'm looking to adapt the approach here: https://github.com/joefitzgerald/go-runtime/pull/1. An open question for me is: given `GO15VENDOREXPERIMENT` and the change to use `vendor` and the tools' awareness of `vendor` (which is still ongoing; some still need updates), do we need to use `godep` to determine the `GOPATH` or can we rely on the convention that `GO15VENDOREXPERIMENT` establishes (that I assume will be shipped by default in 1.6)?

joefitzgerald:
See also: https://github.com/tools/godep/pull/244

cp16net:
I was just thinking this would be useful and looks like it might not have been worked on in a while. While using the `./Godeps/_workspace` would be useful if `GO15VENDOREXPERIMENT` is set to add the godeps then `./vendor` would be set. I believe the full path is required to the directory as well for the `GOPATH` and not a relative directory. 

my 2 cents.

joefitzgerald:
This seems to no longer be useful, due to the vendor experiment flag being removed in 1.7, and the general use of `vendor` instead of `Godeps` as a folder for vendored dependencies. On this basis, I am closing the PR but I do really thank you for your contributions here.

PR #212, opened 2015-06-22
Remove autocomplete-plus install from CI scripts.

zmb3:
This package is now bundled with Atom and doesn't need to be installed separately.

Resolves the following build warning:

>The autocomplete-plus package is bundled with Atom and should not be explicitly installed.
You can run `apm uninstall autocomplete-plus` to uninstall it and then the version bundled
with Atom will be used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,6 @@ notifications:
     on_success: never
     on_failure: change
 
-env:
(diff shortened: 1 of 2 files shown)


joefitzgerald:
Thanks!

PR #227, opened 2015-06-28
Fix issue were godef didn't set GOPATH

miguel250:
The` GOPATH` set in the package settings is not being use when `godef` is run.


joefitzgerald:
Thanks!

PR #234, opened 2015-07-08
Fix godef command on Windows (fixes #221)

### LINKED ISSUE #221: split result by executing godef command wrong on windows
in godef.coffee:59
```coffee
outputs = stdout.split(':')
targetFilePath = outputs[0]
      unless fs.existsSync(targetFilePath)
        @bailWithWarning("godef suggested a file path (\"#{targetFilePath}\") that does not exist)", callback)
        return
row = parseInt(outputs[1], 10) - 1
col = parseInt(outputs[2], 10) - 1
```
if i found the definition of `RWMutex` by godef, t got `D:\go\src\sync\rwmutex.go:18:6`
The outputs is ['D', '\go\src\sync\rwmutex.go', '18', '6'] by splitting stdout with ':'.

ma314smith:
Just a small change to handle the first colon in the windows file path (ex. C:\).

### DIFF
--- a/lib/godef.coffee
+++ b/lib/godef.coffee
@@ -48,7 +48,7 @@ class Godef
     @reset(@editor)
     @gotoDefinitionForWord(word, done)
 
-  gotoDefinitionForWord: (word, callback = -> ) ->
+  gotoDefinitionForWord: (word, callback = -> undefined) ->
     message = null
     done = (exitcode, stdout, stderr, messages) =>
       unless exitcode is 0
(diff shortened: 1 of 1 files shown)


joefitzgerald:
Nice - thanks! Looks like a couple of minor lint errors. Do you want to fix? I will get to it in an hour otherwise...

ma314smith:
I'm not sure if you get notifications on additional commits to the PR, so just commenting to let you know I resolved the lint issues.

PR #239, opened 2015-07-15
updated godef package location

ma314smith:
The godef package has been moved to github (see http://godoc.org/code.google.com/p/rog-go/exp/cmd/godef).  This change will allow Get Missing Tools to install it.


joefitzgerald:
LGTM. Thanks!

PR #247, opened 2015-07-30
Add goimports  as a command(cmd-i)

timesking:
It's handy! goimports on save always remove some package I would use.
Please consider merge this request.

### DIFF
--- a/lib/dispatch.coffee
+++ b/lib/dispatch.coffee
@@ -9,6 +9,7 @@ Executor = require('./executor')
 Environment = require('./environment')
 GoExecutable = require('./goexecutable')
 Godef = require('./godef')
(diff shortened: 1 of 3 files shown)


joefitzgerald:
goimports is already a part of gofmt.coffee - it's also the default configuration for the format tool. What does this let you do that the current code does not? Run gofmt on save but run goimports as a command?

timesking:
goimports is useful tools, but if it is triggered in on-save, it is a little bit annoying when coding golang. auto run goimport in saving, it may remove the package I would like to use.

joefitzgerald:
This is now possible via the command palette in go-plus >= 4.0.0.

PR #250, opened 2015-08-02
Adding support for quick imports with 'go list ...'

ChrisTrenkamp:
Adding support for quickly adding an import in a Go source file.  It uses the command "go list ..." to create a package list and lists the packages in a dropdown.  The default keymap is ctrl+alt+i.

### DIFF
--- a/lib/golist.coffee
+++ b/lib/golist.coffee
@@ -130,7 +130,7 @@ class GoList extends SelectListView
   getNewl: (str) ->
     elF = (el) ->
       return el is '\r\n'
-    newlines = (str.match(/(?:\r?\n)/g) || [])
+    newlines = (str.match(/(?:\r?\n)/g) or [])
     crlf = newlines.filter(elF).length
     lf = newlines.length - crlf
     if crlf > lf


joefitzgerald:
What would this look like if - instead of having a UI for this in go-plus - you created a special-purpose autocomplete-plus provider (inside go-plus) that only kicked in when the scope for the cursor was inside an import statement?

This would allow us to have approximately the same functionality without having to maintain a UI. Thoughts?

ChrisTrenkamp:
I'm coming from the Sublime editor, which has the GoSublime plugin.  I miss being able to quickly import packages without losing my position. Creating a special autocomplete for import statements would have the same functionality, but not quite the convenience.

I think being able to quickly bring up a UI and selecting the package you want would be more convenient than scrolling all the way back up in the source file, importing the package, and scrolling back to where you were.

With that said, originally, I was going to make this a separate package, but when I got to the point where I needed to add a setting for adding the Go executable to the path, I thought it would make more sense to add this to the go-plus package.  If maintenance is a concern, I can keep this as a separate package.

tuvistavie:
I also come from Sublime and was looking for this feature.
I also agree that a quick list is more handy than an autocomplete-plus provider, which would
force the user to go to the import list.

holys:
ping

marconi:
+1

joefitzgerald:
This is fixed via #720.

PR #261, opened 2015-08-24
Better godef integration: local defs, imports

greensnark:
go-plus' current godef integration is limited to top-level and exported names,
and does not work for method calls:

    buf := bytes.Buffer{}
    buf.Write([]byte{0x0}) // godef on Write and buf didn't work.

This is because godef was being run as:

    godef -f filebeingedited.go buf.Write

which doesn't give godef enough context to know what "buf" and "Write" are.

The fix is to run godef with the offset (-o) option to give it the offset
in the file where the cursor is, and let godef figure out what the user's
trying to navigate to.

This commit also keeps track of godef navigation on a stack and allows going
back to previous locations where godef was triggered with the godef-return
command.

### DIFF
--- a/keymaps/go-plus.cson
+++ b/keymaps/go-plus.cson
@@ -3,4 +3,6 @@
 'atom-text-editor[data-grammar="source go"]:not(.mini)':
   'ctrl-alt-c': 'golang:gocover'
   'ctrl-alt-shift-c': 'golang:cleargocover'
+  # Shouldn't these be ctrl-alt for portability?
   'alt-cmd-g': 'golang:godef'
+  'alt-cmd-shift-g': 'golang:godef-return'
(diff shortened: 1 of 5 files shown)


joefitzgerald:
LGTM. Thanks, this is great!

keymaps/go-plus.cson
```diff
@@ -3,4 +3,6 @@
 'atom-text-editor[data-grammar="source go"]:not(.mini)':
   'ctrl-alt-c': 'golang:gocover'
   'ctrl-alt-shift-c': 'golang:cleargocover'
+  # Shouldn't these be ctrl-alt for portability?
```
joefitzgerald: Yes, you're right.

lib/executor.coffee
```diff
@@ -85,5 +85,5 @@ class Executor
       err.handle()
       callback(127, output, error, messages)
 
-    if input
+    if input?
```
joefitzgerald: Thanks for catching this.

PR #265, opened 2015-08-26
godef: handle multibyte UTF-8 codepoints

greensnark:
My previous update to use godef's -o option passed in UTF-16 (Javascript
string) offsets to godef, which wants byte offsets in the source file. Give
godef the UTF-8 offset instead (since Go requires that source files must be
encoded in UTF-8).

(Switched to Buffer.byteLength to save allocating a new Buffer that's then immediately thrown away)

### DIFF
--- a/lib/godef.coffee
+++ b/lib/godef.coffee
@@ -52,10 +52,12 @@ class Godef
       @bailWithWarning('Godef only works with a single cursor', done)
       return
 
-    editorCursorOffset = (e) ->
-      e.getBuffer().characterIndexForPosition(e.getCursorBufferPosition())
+    editorCursorUTF8Offset = (e) ->
+      characterOffset = e.getBuffer().characterIndexForPosition(e.getCursorBufferPosition())
+      text = e.getText().substring(0, characterOffset)
+      new Buffer(text, "utf8").length
 
-    offset = editorCursorOffset(@editor)
+    offset = editorCursorUTF8Offset(@editor)
     @reset(@editor)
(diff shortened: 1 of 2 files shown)


greensnark:
AppVeyor build seems to be failing because of unrelated cgo issues on Windows (looks like Go 1.5 is tripping over .c files from the Go 1.4 runtime package?)

joefitzgerald:
Hmm: https://github.com/golang/go/issues/9410

joefitzgerald:
Also: http://stackoverflow.com/questions/32115229/go-1-5-error-imports-runtime-c-source-files-not-allowed-when-not-using-cgo-o - perhaps I need to forcefully delete the old go installation on AppVeyor.

greensnark:
Yes, sounds like AppVeyor still has old Go 1.4 installed somewhere...

zmb3:
I'd be interested in hearing if you find a more efficient way to do the character->byte conversion.  I ran into the same thing with the `gorename` integration.

In my case, I actually submitted a change to gorename to accept a character offset (which is still pending review).  It's much easier, IMO, and probably significantly faster, to do the conversion in Go code rather than in Atom.

greensnark:
Is there a convenient way to get true character offsets in Atom? As far as I know it's easy to get a UTF-16 code unit offset, but that isn't the same as a character offset for non-BMP Unicode. For instance, if I open a new Atom editor pane and type just "𝒞", the cursor position at the end of the "𝒞" claims to be 1:3, when it really should be 1:2.

In that context, I think using the byte offset notion isn't so bad after all :)

PR #276, opened 2015-09-08
Add support for emoji fixes #275

### LINKED ISSUE #275: Autocomplete stops working when source file contains emoji
I'll create a concise testcase for this when I have some free time but I'm noticing that autocomplete stops working when I have emoji in the file (it never filters the list - always displays all of the options).

danielchatfield:


danielchatfield:
A single emoji counts as 2 in the characterIndex, the atom docs for this method could be more explicit.

joefitzgerald:
So in essence, the method isn't UTF-8 compatible. Got it. My question then is, can we still stick with the `c` prefix in our offset, and figure out the actual number of UTF-8 characters (rather than the byte offset)?

danielchatfield:
That would be one way of doing it but I'm not sure it would be preferable. gocode converts a character offset into a byte offset anyway. Byte offsets are unambiguous and _just work_.

PR #278, opened 2015-09-10
added support for GO15VENDOREXPERIMENT

interlock:
Title says it all. Has one spec to validate the env is set.

### DIFF
--- a/lib/go-plus.coffee
+++ b/lib/go-plus.coffee
@@ -61,48 +61,54 @@ module.exports =
       type: 'boolean'
       default: true
       order: 10
+    vendorExperiement:
+      title: 'Enable GO15VENDOREXPERIMENT'
+      description: 'Enable support for Go 1.5 Vendor experiement'
+      type: 'boolean'
+      default: false
+      order: 11
     vetOnSave:
       title: 'Run Vet Tool On Save'
       description: 'Run `go vet` each time a file is saved'
       type: 'boolean'
       default: true
-      order: 11
+      order: 12
     vetArgs:
       title: 'Vet Arguments'
       description: 'Arguments to pass to `go vet` (these are not usually needed)'
       type: 'string'
       default: ''
-      order: 12
+      order: 13
     getMissingTools:
       title: 'Automatically Get Missing Tools'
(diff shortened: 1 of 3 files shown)


interlock:
Would love to see this integrated. Anything I can do to help it along?

joefitzgerald:
Given that you should be launching atom from your terminal, this may be redundant. Is there an issue with either:

* Set the environment variable `GO15VENDOREXPERIMENT` and then launch `atom` from the terminal: `$ atom .`
* Launch `atom`, setting the environment variable for the process: `$ GO15VENDOREXPERIMENT=1 atom .`

screen shot 2015-09-15 at 3 36 47 pm

joefitzgerald:
(By the way, thank you so much for the contribution and the interest in the project!!!)

interlock:
@joefitzgerald I launch from the CLI with `atom` and my env has `GO15VENDOREXPERIMENT` set and confirmed with `export|grep GO15VENDOREXPERIMENT`

The executer appears to trim down the environment at some point and it is now included in the exec calls. This adds it. Most of the tools being called could benefit from this, but so far just the one has it.

Quentin-M:
Most tools should also ignore the `vendor` directory, otherwise everything become really, really slow.

interlock:
@Quentin-M What's your experience with tools getting slow that do not ignore `vendor`?

So far no noticeable difference, but when you switch to using `vendor` you tend to keep your GOPATH much cleaner. Not many more extra packages to scan in that case.

@joefitzgerald Any comment on why GO15VENDOREXPERIMENT env variables are not getting passed to the tools via go-plus? When I traced it, Atom's execute wrapper was creating a new env that did not include the launch environment variables. Curious to know if you have experienced something different, or were working on an assumption that the shell env does make its way to the executor.

joefitzgerald:
@interlock I really don't know why that's happening for you. It shouldn't be. Would you be open to a screen sharing session? If so, follow https://twitter.com/joefitzgerald and we can exchange contact details via DM.

joefitzgerald:
The vendor experiment is fully supported so closing this out. Thank you so much for taking the time to contribute!

Issue #213, opened 2015-06-23
Switch To Use gometalinter


soygul:
https://github.com/alecthomas/gometalinter

Just seeking to understand if this is a popular tool with the community. Given that it normalizes output from all known linters, can make linter editor integrations less manual.

joefitzgerald:
Indeed. I think gometalinter is great. I am going to be revisiting the way we run all the tools we use and integrate with https://github.com/atomlinter/linter. As part of this, I will look into the use of gometalinter.

soygul:
:+1:

rtyer:
:+1:

adamclerk:
:thumbsup:

santagada:
Great, have messages integrated with the line of the problem is way better than having them on a tool box, and the metalinter is the way to go for performance and completeness reasons.

Can other people send pull requests for this or is anyone active in fixing this?

steelbrain:
For the DIY people on here, it's super-simple to write a linter, see https://github.com/atom-community/linter/wiki/Linter-API and https://github.com/atom-community/linter/wiki/Migrating-to-the-new-API

joefitzgerald:
This is pending a rewrite of the core runtime detection in https://github.com/joefitzgerald/go-runtime, so that we don't end up with a litany of different ways this is done. While it's trivial to write a linter it is very non trivial to do so in a way that a user has a zero configuration experience based on their current environment. go-runtime will also pave the way for gb and the GO15VENDOREXPERIMENT.

santagada:
Thank you @steelbrain I will look into that. @joefitzgerald go-runtime doesn't look very promising seeing that it hasn't been any commits for 5 months right? Is there anything we can help you with that then?

joefitzgerald:
@santagada it's because I haven't pushed the code that I have locally. For a period of time I was blocked on waiting for package-sets to be included in core (as I was the committer) but recently I decided to move ahead without package-sets as its such a large piece of work it will take time to ship. I will push new code there tomorrow, and I'd welcome the assistance.

santagada:
@joefitzgerald any news?

mishak87:
Any progress?
Linter API evolved quite a bit and maybe writing own package for go linting would be best way to go.

joefitzgerald:
Yep. https://atom.io/packages/gometalinter-linter.

mishak87:
@joefitzgerald Sorry I must have missed it :smile:. I promise I will do lot of user testing.

[...]

PR #284, opened 2015-09-25
Linter support: https://atom.io/packages/linter

greensnark:
Make go-plus an atom-linter provider for warning and error messages. This is controlled by the new useAtomLinter setting which defaults to true.

(atom-linter provides gutter annotations and navigation between errors/warnings, and is actively maintained.)

The atom linter option is set to true by default, and will automatically be preferred to atom-message-panel if both are available.


greensnark:
Not sure why AppVeyor build is failing...

zmb3:
I think Joe's been planning to use atom-linter for a while (as part of #213) but there's some work re: runtime detection that still needs to be done.

greensnark:
@zmb3 Thanks for the word, I'll leave this up and hope it doesn't duplicate work Joe's already done.

greensnark:
The AppVeyor build failure is pretty interesting: the atom process that `apm test` spawns is exiting with code -1073741819 which is suspiciously identical to 0xC0000005, viz. Windows' Access Violation / segfault. I can reproduce this on a Windows VM with Atom 1.0.19. The crash seems unrelated to this pull request since I can replicate it easily with many other atom packages.

sztanpet:
may i also suggest using gometalinter as the linter? it has so many useful tools including all the ones already included

joefitzgerald:
Update: https://atom.io/packages/gometalinter-linter.

joefitzgerald:
Hey @greensnark thanks so much for the contribution! It helped me significantly in my implementation of gometalinter-linter.

PR #287, opened 2015-10-01
Add fuzzy matching to autocomplete suggestions

tylerb:
This PR adds fuzzy autocomplete matching using [fuzzaldrin](https://github.com/atom/fuzzaldrin). Fuzzy matching is disabled by default, but can be enabled using `fuzzyAutocomplete: true` in the config file.

There is room for improvement, such as ordering suggestions based on frequency of usage. Other improvements may also be possible, but my knowledge in this area is limited. This is a naive approach that simply pipes everything through fuzzaldrin.

Please note: I am rather new to coffeescript and this is my first time working on an Atom package. I'm sure there is room for improvement in my code. Feedback would be most appreciated. I'm happy to clean/alter things to make this PR acceptable.

Thank you!


dynamix:
exactly what I was look for! However it doesn't work properly (just tried it). Sometimes it "eats" the dot for me  and no suggestions show up before you type the first letter after the dot. If I have some time later, I'll give it a try to fix it.

joefitzgerald:
`gocode` provides our matches, and provides a list of values that is more comprehensive than the list of values resulting from the use of this PR. For that reason, I'm going to close this. Thanks so much for your contribution; many apologies for leaving this PR open for so long!

PR #330, opened 2015-12-18
add "working" message panel

TrueFurby:
For me it sometimes takes little while to run the tools and show results so I added info message panel before running tools to show that it's running them.


joefitzgerald:
The message panel is no longer part of the package, so closing this out. Thanks so much for contributing!

PR #485, opened 2016-10-12
Merge gofmt back into go-plus

zmb3:
- [x] Add formatter and specs from gofmt package
- [x] Add menu items and keymaps
- [x] Group gofmt config settings under a separate object
- [x] Add code to remove the gofmt package which will no longer be necessary

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,8 @@ install:
   - source "$NVM_DIR/nvm.sh"
   - nvm install node
   - nvm use node
+  - go get -u golang.org/x/tools/cmd/goimports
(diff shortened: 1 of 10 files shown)


zmb3:
@joefitzgerald please review.

Once we start merging other packages that perform some action on save we should update so that we're only subscribing to save events in one place and then delegating to other tools.  I don't think that's necessary prior to merging this though.

joefitzgerald:
> Once we start merging other packages that perform some action on save we should update so that we're only subscribing to save events in one place and then delegating to other tools

Agreed...

PR #487, opened 2016-10-14
Consolidate builder-go into go-plus

zmb3:
This will pave the way for better integration between go-plus features.  For example we'll be able to skip linting if the build fails.
- [x] add builder
- [x] add specs
- [x] uninstall builder-go package
- [x] update readme
- [ ] remove linter provider and integrate with go-plus panel


zmb3:
@joefitzgerald thoughts on merging this and doing messages panel later?

PR #488, opened 2016-10-15
[WIP] Vertical panel

simurai:
This adds an option for a vertical panel.

screen shot 2016-10-15 at 12 55 21 pm
- [x] Add vertical panel styles
- [ ] Add config option to switch between horizontal/vertical
- [ ] Make it resizable vertically


simurai:
@joefitzgerald 

The styles can be enabled by adding a `is-vertical` class to `go-plus-panel`. Plus change the inline styles from `height` to `width`.

```html
<atom-panel class="go-plus-panel is-vertical" style="width: 25vw; min-width: 15vw;
  max-width: 50vw;">
</atom-panel>
```

lloiser:
Do you need any help with this?

joefitzgerald:
screen shot 2016-11-06 at 3 41 03 pm

PR #494, opened 2016-10-27
Always show stderr output

Andrzej:
Currently stderr output is not always visible. I feel it is always useful to see it, especially when tests fail.

In any case, thank you for awesome plugin!


joefitzgerald:
LGTM. Thanks!

Issue #490, opened 2016-10-17
Only enable these plugins in go files


rumpelsepp:
Is it possible to enable these plugins only for go files? I often edit files in multiple languages; these go commands are useless in, let's say, python files. My atom starts to get bloated because every package is enabled in every file. :/

joefitzgerald:
Closing as duplicate of #77.

Issue #77, opened 2014-10-15
Do not load if no .go files are being currently in treeview / editor


ptomasroos:
I think its strange that this plugin bootstraps even though i'm doing node sometimes.

What would be a good approach to defer the loading of this when its only about to be used?

jokeyrhyme:
I feel lazy-loaded plugins that are specific to certain file types is a job that Atom itself should be doing, instead of requiring plugin authors to all implement this sort of detection themselves.

joefitzgerald:
I actually added the features to Atom to permit this to function correctly (activating a package on first use of a grammar). It's just not been done yet; there are a number additional changes I need to make to the package to enable it, which are in progress currently.

ptomasroos:
WOW! 💃

joefitzgerald:
This is released in [`v5.0.6`](https://github.com/joefitzgerald/go-plus/releases/tag/v5.0.6).

Issue #51, opened 2014-08-07
Hide panel at startup


alexcesaro:
I think startup info (`Using Go: go1.2.1 linux/amd64 (@/usr/bin/go)...`) should be hidden at startup. Especially when there is no error and `Show Panel When No Issues Exist` is unchecked.

PS: Thanks for this nice plugin!

jamescmartinez:
@joefitzgerald What's the update on this? Can I help in any way?

gempir:
Any progess on this?

joefitzgerald:
Yep update to 4.x; the panel is gone.

PR #499, opened 2016-11-03
panelDisplayMode defaults to "closed"

jokeyrhyme:
I use Atom mostly for JavaScript and CommonMark files.

Even though I find go-plus a pleasure to use when I'm writing Go code, I'd prefer it if the default state of the panel (for new Atom windows) was "closed".

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
       "title": "Panel Display Mode",
       "description": "Control the behavior of the go-plus panel when a new Atom window is opened",
       "type": "string",
-      "default": "open",
+      "default": "closed",
       "enum": [
         "open",
         "closed"


zmb3:
You can get the behavior you want by changing the panel display mode to closed in the go-plus settings.

I believe there's an issue for hiding the panel by default until a *.go file is open, which to me seems like a better solution for people that don't work exclusively with Go.  I think if the panel is hidden by default some users won't even know it exists.

jokeyrhyme:
> You can get the behavior you want by changing the panel display mode to closed in the go-plus settings.

I've been using that setting for a while now. I just reinstalled Atom on a fresh machine again and figured the default behaviour was ever so slightly unnecessary.

> I believe there's an issue for hiding the panel by default until a *.go file is open

There's #77, and I think even #51 is loosely related to this.

> I think if the panel is hidden by default some users won't even know it exists.

There is the clickable "go-plus" indicator in the bottom status bar area. Even someone that lives 100% in Go probably won't need the window open all the time, and would probably appreciate dedicating more screen area to the code window by default.

I'd suggest that my proposed new default setting here is a good compromise until either #77 is implemented or some sort of "show panel when saving a *.go file triggers tests" functionality is available. I'm not the boss of anyone, of course. Your call. :)

zmb3:
There's also #490.

>Even someone that lives 100% in Go probably won't need the window open all the time

Totally agree.

> I'm not the boss of anyone, of course. Your call. :)

I'm no boss either :-)  Just trying to gauge the value of this proposal.  It saves you from having to change a setting when you set up Atom on a new machine...but I suspect this is probably not frequently, and you already have to set up and install all of your packages in this case anyway.

It would be neat if Atom provided an API for packages to detect the first time they are activated after an installation (or even update).  This would allow you to customize the "first time experience" and do things like point out UI elements, check that the user's environment is sane, prompt the user to set up their defaults, etc.  Maybe we could explore something along this avenue.

joefitzgerald:
Given that:

1. we intend the default to remain `open`
1. we intend to defer the initialization of the panel until the first use of a `.go` file or a go-plus command
1. there exists a way to configure the panel to be `closed` on startup

I am going to close this. It's not worth the churn of changing the defaults for what may be a short period of time before we complete 2., above.

Many thanks for your contribution!

jokeyrhyme:
Works for me. :)

PR #503, opened 2016-11-15
Consolidate Downstream Packages Into go-plus

### LINKED ISSUE #498: Right side panel
When reading godoc docs, it would be helpful to have the panel on the right because the docs are often too long for the bottom panel and require scrolling. This works better with multiple monitors or very wide (21:9) monitors because vertical space doesn't increase but horizontal space is increasing. I propose 2 different solutions: 

1) provide an option to have the panel on the right
2) create a second panel for docs only on the right -- it could be html rendered because it wouldn't be shared with all the other tools that need plain text output

I prefer 2, but 1 would be sufficient.

joefitzgerald:
The verdict is in: the experiment of decomposing go-plus into a collection of loosely coupled packages collaborating via services was not successful. It resulted in multiple levels of indirection, hampering the ability for people to:

* report issues
* find package settings
* understand the code base
* contribute to the package

This PR collapses the world back down into a single package: go-plus. It preserves the services that are provided for `go-config` and `go-get`, but I strongly encourage package authors making use of those packages to stop using `package-deps` to install `go-plus` (or the old `go-config` and `go-get` packages); it introduces a brittle package ecosystem, when changes like this are made.

* [x] Activate package via the `core:loaded-shell-environment` `activationHook` to ensure a patched environment is available for users of OS X / Linux in some launch scenarios (e.g. Dock / Unity Launcher)
* [x] Allow the go-plus panel to have `vertical` or `horizontal` orientations
* [x] Consolidate package configuration and arrange it into related sections
* [x] Consolidate `go-config`
* [x] Consolidate `go-get`
* [x] Consolidate `builder-go`
* [x] Consolidate `tester-go`
* [x] Consolidate `gofmt`
* [x] Consolidate `autocomplete-go`
* [x] Consolidate `gometalinter-linter`
* [x] Consolidate `navigator-go`
* [x] Work with @zmb3 to consolidate [`godoc`](https://github.com/zmb3/godoc)
* [x] Work with @zmb3 to consolidate [`gorename`](https://github.com/zmb3/gorename)
* [x] Work with @zheng1 to consolidate [`go-hyperclick`](https://github.com/zheng1/go-hyperclick)
* [x] Ensure `gometalinter --install` is run after `golang:update-tools`
* [x] Add a timeout to executor, to prevent runaway processes
* [x] Add configurable timeout for tests
* [x] Add configurable timeout for go get
* [x] Switch default for go to definition from godef to guru

- Fixes #487 
- Fixes #488 
- Fixes #491 
- Fixes #492 
- Fixes #495 
- Fixes #497 
- Fixes #498 
- Fixes https://github.com/joefitzgerald/tester-go/pull/36

<hr />

The following additional features will be added in subsequent PRs:

* Add option for go to definition that makes use of gogetdoc
* Add ability for the hyperclick provider to display godoc as an overlay on hover
* Allow coverage display style to be one of: `gutter`, `highlight`, `both`, adopting code from https://github.com/joefitzgerald/tester-go/pull/36


lloiser:
@jnmoal please keep track of this merge request so you can update [go-guru](https://github.com/jnmoal/go-guru) after this has been merged!

lib/rename/gorename.js
```diff
@@ -0,0 +1,181 @@
...
+      return {success: true, result: r}
+    })
+  }
+
+  checkForTool (editor) {
+    if (!this.goconfig) {
+      return Promise.resolve(false)
+    }
+
+    const options = {}
```
lloiser: looks like `getLocatorOptions`

lib/components/godoc-panel-view.js
```diff
@@ -0,0 +1,39 @@
+/** @babel */
+/** @jsx etch.dom */
+
+import etch from 'etch'
+
+export default class GodocPanelView {
```
zmb3: I could be talked into this.  I don't particularly like inheritance if it isn't necessary, so when this was in my own package I opted not to create the superclass.

https://github.com/atom/etch#what-about-a-component-superclass

lib/build/builder.js
```diff
@@ -0,0 +1,215 @@
...
+        }
+      })
+      return { output: r.stderr.trim(), linterName: 'test' }
+    }).catch((e) => {
+      console.log(e)
+      return { output: '', linterName: 'test' }
+    })
+  }
+
+  getLocatorOptions (editor = atom.workspace.getActiveTextEditor()) {
```
zmb3: I looked at these but they aren't identical - there are little differences all over the place and I didn't want to break something subtle that would hold up a release.

lib/navigator/godef.js
```diff
@@ -0,0 +1,304 @@
...
+        dismissable: true,
+        icon: 'location',
+        detail: 'godef returned invalid file path',
+        description: loc.filepath
+      })
+      return false
+    })
+  }
+
+  stat (loc) {
```
zmb3: Good catch!

lib/components/godoc-panel-view.js
lloiser: They are talking about the `etch` package and why they don't want to add a superclass over there. But having one here is ok. It is already used by the other panel views so it just feels weird to not use it for this one too

lib/build/builder.js
lloiser: ok so I quickly searched for `getLocatorOptions` and compared them: 
* `godef`, `linter`, `get-manager` and `builder` use the same (linter has some additional safety `if`s)
* `formatter`, `information` and `tester` use nearly the same version although `formatter` has some relative path handling in `projectPath` that could affect the others... 🤔

lib/components/godoc-panel-view.js
lloiser: btw: I found the problem why after some time the panel does not work anymore (as you mentioned in the commit message 4227b85). Because etch destroys the DOM node. This is actively prevented by `EtchComponent.destroy` but not in your `GodocPanelView`, so inheriting from `EtchComponent` would solve this problem too 😉

zmb3: I concede! 👍

lib/autocomplete/gocodeprovider.js
```diff
@@ -0,0 +1,634 @@
...
+
+    if (this.subscribers && this.subscribers.length > 0) {
+      for (const subscriber of this.subscribers) {
+        subscriber(p)
+      }
+    }
+    return p
+  }
+
+  onDidGetSuggestions (s) {
```
zmb3: Question for @joefitzgerald - it might be a callback for AC+

lib/main.js
```diff
@@ -163,45 +270,117 @@ export default {
 
   consumeViewProvider (provider) {
     if (!provider || !provider.view || !provider.model || !provider.model.key) {
+      return new Disposable()
+    }
+    const view = provider.view
+    const model = provider.model
+    if (model.key && (model.key === 'go' || model.key === 'test' || model.key === 'reference')) {
```
zmb3: is that true?

lib/main.js
```diff
@@ -163,45 +270,117 @@ export default {
...
+    const {Godef} = require('./navigator/godef')
+    this.godef = new Godef(this.getGoconfig())
+    this.subscriptions.add(this.godef)
+    return this.godef
   },
 
-  registerTools () {
-    if (this.toolsRegistered || !this.goget) {
+  provideLinter () {
+    return this.getBuilder()
```
zmb3: @joefitzgerald ?

lib/main.js
```diff
@@ -1,81 +1,191 @@
...
+const oldPackages = [
+  'gofmt',
+  'tester-go',
+  'builder-go',
+  'autocomplete-go',
+  'godoc',
+  'gorename',
+  'go-hyperclick',
+  'navigator-go',
+  'go-get',
```
zmb3: I'm thinking `go-get` and `go-config` should not be in this list as they will still exist for other plugins (like `go-signature-statusbar` or `go-debug`)

joefitzgerald: No, go-get and go-config will be end of life once this PR merges. Packages wishing to make use of that behavior will use go-plus as their dependency, as it offers those same services.

lib/main.js
joefitzgerald: You're right, we need a second provide function.

lib/lint/linter.js
```diff
@@ -0,0 +1,179 @@
...
+  lint (editor) {
+    if (!this.ready() || !editor) {
+      return []
+    }
+
+    const buffer = editor.getBuffer()
+    if (!buffer) {
+      return []
+    }
+    let args = atom.config.get('go-plus.gometalinter.args')
```
lloiser: it's `go-plus.lint.args` instead of `go-plus.gometalinter.args`

README.md
```diff
@@ -45,14 +63,10 @@ go get -u github.com/rogpeppe/godef
 
 You can file any `go-plus` issues [here](https://github.com/joefitzgerald/go-plus/issues/new). You can optionally file an issue in a downstream repository for anything related to a bundled package:
 
-* [`runtime detection`](https://github.com/joefitzgerald/go-config): [create issue](https://github.com/joefitzgerald/go-config/issues/new)
```
lloiser: Clean up: only `go-debug` will be installed, everything else is already part of this package.

README.md
```diff
@@ -11,17 +11,34 @@ This package includes the following functionality:
...
+* Run `go install .` and `go test -c -o {tempdir} .` to verify your code can compile and to keep `gocode` suggestions up to date
 * Run tests, display test output, and display test coverage using `go test -coverprofile`
+* Display documentation for identifiers in source code using
+  [`gogetdoc`](https://github.com/zmb3/gogetdoc)
+* Rename the symbol under your cursor using `gorename`
 
 This package adds extra functionality by installing the following additional packages:
 
 * [autocomplete-go](https://atom.io/packages/autocomplete-go): Autocomplete using `gocode`
-* [builder-go](https://atom.io/packages/builder-go): Run `go install .` and `go test -c -o {tempdir} .` to verify your code can compile and to keep gocode suggestions up to date
```
lloiser: Clean up: only `go-debug` will be installed, everything else is already part of this package.

lib/lint/linter.js
zmb3: @lloiser thanks for catching all of these errors

you have push access right?  why not just push fixes when you spot something wrong?  that way we won't forget to go back to any of the things you've pointed out.

lib/autocomplete/gocodeprovider.js
joefitzgerald: This is used in the specs.

lib/test/test-panel-manager.js
```diff
@@ -63,7 +73,7 @@ class TestPanelManager {
       this.view.update()
     }
 
-    if (this.props.exitcode && this.props.exitcode === 1 && this.requestFocus) {
+    if (this.props.exitcode && this.props.exitcode !== 1 && this.requestFocus) {
```
lloiser: why `exitcode !== 1`?

PR #506, opened 2016-11-18
Use activationHooks to speedup editor startup time

sadovnychyi:
From the [docs](http://flight-manual.atom.io/hacking-atom/sections/package-word-count/):

> `activationHooks`: an Array of Strings identifying hooks that trigger your package's activation. The loading of your package is delayed until one of these hooks are triggered. Currently, the only activation hook is `language-package-name:grammar-used` (e.g., `language-javascript:grammar-used`)

This should speed up Atom startup time while no go files presented (since most of people use Atom as universal editor for many languages).

This should also hide go-plus panel while I'm working on python files for example. It should become visible after I open any go-file.

Not sure if specs will work correctly after this though.

#505

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,10 @@
   "bugs": {
     "url": "https://github.com/joefitzgerald/go-plus/issues"
   },
-  "activationHooks": ["core:loaded-shell-environment"],
+  "activationHooks": [
+    "core:loaded-shell-environment",
+    "language-go:grammar-used"
+  ],
   "activationCommands": {
     "atom-workspace": [
       "golang:get-package",


joefitzgerald:
Just for context, I was the person who added this to Atom core 🤓. So naturally I'm very aware of it and aspire to use it (that's why I added it). 

Right now we are dealing with a tactical issue related to the environment. I may need to expand activationHooks to allow specification for activation only when two hooks have triggered.

sadovnychyi:
Good to know! Looking forward to it. Seems like after #503 all commands will be context aware which I'm very happy to see.

PR #515, opened 2016-12-19
Ensure go-plus Panel Does Not Show On Startup, If So Configured

### LINKED ISSUE #514: The Panel options are broken
The panel will be open always when I start Atom, regardless of whether I select it to be open or closed.

joefitzgerald:
- Fixes #514

### DIFF
--- a/lib/panel-manager.js
+++ b/lib/panel-manager.js
@@ -12,18 +12,22 @@ class PanelManager {
     this.subscriptions = new CompositeDisposable()
     this.viewProviders = new Map()
     this.resizeSubscriptions = new CompositeDisposable()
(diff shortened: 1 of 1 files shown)


lib/panel-manager.js
```diff
@@ -12,18 +12,22 @@ class PanelManager {
     this.subscriptions = new CompositeDisposable()
     this.viewProviders = new Map()
     this.resizeSubscriptions = new CompositeDisposable()
+    this.initialized = false
     this.subscribeToConfig()
```
lloiser: yet your approach feels quite solid too. Just wanted to mention the actual cause...

PR #520, opened 2016-12-22
adds the ability to define a custom directory for test and coverage runner

xpressivecode:


joefitzgerald:
What's the motivator for this?

xpressivecode:
There was a question in slack. It seems that some individuals restrict exec access to their tmp dir. Either by choice or as a company policy. Thought it might help, vs always having to run something like ```TMPDIR=/custom atom .```

joefitzgerald:
It seems this only happens on certain operating systems. Maybe we could make this automatic in some way?

xpressivecode:
It looked like there was some debate in regards to where build artifacts should end up. So automating it looked like it may upset a few people, vs letting them decide. Maybe updating the FAQ to suggest running ```TMPDIR=/custom atom .``` and removing this pr is enough instead?

joefitzgerald:
Opened #722 to document `TMPDIR=/custom atom .`.

PR #557, opened 2017-01-24
Improve navigation with guru

lloiser:
It currently fails if the cursor is at the end of a word:
```go
fmt.Println ("hello world")
//         ↑ if cursor is here between `ln` and `(` it fails
```

Solution: call guru with `position.column - 1` (simply step back 1 character)

### DIFF
--- a/lib/navigator/godef.js
+++ b/lib/navigator/godef.js
@@ -137,31 +137,35 @@ class Godef {
         options.input = archive
       }
 
-      const goto = (pos) => {
-        const offset = this.utf8OffsetForBufferPosition(pos, editor)
-        const args = ['-json', 'definition', filepath + ':#' + offset]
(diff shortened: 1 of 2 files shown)


joefitzgerald:
Wouldn't it be more efficient to just detect the end of a word, and change the position back one char, rather than running / failing / falling back to `pos - 1`?

zmb3:
Agree with Joe, what happens now when the cursor is before the `f` in `fmt`.  It seems you've just shifted the problem from one end of the word to the other.

lloiser:
@joefitzgerald I'll give it a try.

@zmb3 this still works as before. I'm only retrying it if the first guru call fails. Then I go back 1 char and try again.

lloiser:
any objections?

zmb3:
None here.

Issue #308, opened 2015-11-16
Add default build / test flag support


derekperkins:
I have a few different build types depending on my target environment, set using build flags. I'd like to be able to specify a default build flag that gets used for testing. You already have inputs for Format, Lint and Vet arguments, so I'd imagine adding a new input for Build arguments and Test arguments, just like the other three.

joefitzgerald:
Take a peek at the following to see the direction I am headed in:

* [environment](https://atom.io/packages/environment): Gives you a sane environment on OS X even if you forget to launch atom from the terminal (e.g. via Dock, Spotlight, Finder, open, etc.)
* [go-config](https://atom.io/packages/go-config): Provides a core package that all other go packages use to determine the correct go runtime, tools, gopath, etc. for a given file, directory, or project in Atom
* [gometalinter-linter](https://atom.io/packages/gometalinter-linter): Integrates [linter](https://atom.io/packages/linter) with [gometalinter](https://github.com/alecthomas/gometalinter)
* [autocomplete-go](https://atom.io/packages/autocomplete-go): An [autocomplete-plus](https://atom.io/packages/autocomplete-plus) provider that uses [gocode](https://github.com/nsf/gocode)

These are not ready for daily use yet, but do function. I still need to add functionality to fetch tools if missing, support .editorconfig, and automatically install required packages. In order to fully turn go-plus into an "installer" of "packages known to work well together" I still have to address:

* Visualization of test coverage (which really should also show test failures)
* Fetching of required tools
* Migrate go-plus configuration to individual packages
* Create a godef package
* Create a build package

If you want to create the build package, then I'm all for it. You could integrate https://atom.io/packages/build and https://atom.io/packages/go-config to achieve a good outcome?

joefitzgerald:
Related: #281

derekperkins:
I love where you're headed. I don't have time to code that at the moment, but I'm more than happy to test things out and give feedback.

PR #565, opened 2017-02-12
Add config option for additional go test args

### LINKED ISSUE #551: Enhancement: Custom/Additional Test Flags (textbox)
### Description

Justification:

    if glog.V(3) {
    }

Thoughts from others?

zmb3:
Add a new configuration option `go-plus.test.additionalTestFlags` to allow the user to specify additional flags to the `go test` command.

Test flags here take precedence so:
- if this option includes a `-timeout` flag it will override `go-plus.test.timeout`
- if this option includes `-short` or `-v` flags they will be used no matter what the value of the `runTestsWithShortFlag` or  `runTestsWithVerboseFlag` options are (note that these options can still _append_ to the overall set of args)

Fixes #551, See also #308.

The jury's out on whether this actually a good idea or not.  I'm doing minimal validation on what the user specifies, so it's possible for the user to shoot themselves in the foot and specify something that breaks `go test` or overrides our coverprofile settings.

**Note:** it seems our tests are broken on Atom 1.15 beta, which is unrelated to this PR.

### DIFF
--- a/lib/test/tester.js
+++ b/lib/test/tester.js
@@ -228,10 +228,10 @@ class Tester {
 
     let shortFlag = false
     let verboseFlag = false
+    let userSuppliedTimout = false
 
     if (configFlags && configFlags.length) {
       const arr = configFlags.trim().split(/\s+/g)
-      let userSuppliedTimout = false
 
       for (const arg of arr) {
         args.push(arg)
@@ -245,9 +245,10 @@ class Tester {
           verboseFlag = true
         }
       }
-      if (!userSuppliedTimout) {
-        args.push('-timeout=' + timeoutMillis + 'ms')
-      }
+    }
+
+    if (!userSuppliedTimout) {
+      args.push('-timeout=' + timeoutMillis + 'ms')
     }
 
     if (!shortFlag && atom.config.get('go-plus.test.runTestsWithShortFlag')) {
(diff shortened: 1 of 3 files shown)


lib/test/tester.js
```diff
@@ -222,6 +222,44 @@ class Tester {
...
+          userSuppliedTimout = true
+        }
+        if (arg === '-short') {
+          shortFlag = true
+        }
+        if (arg === '-v') {
+          verboseFlag = true
+        }
+      }
+      if (!userSuppliedTimout) {
```
lloiser: the timeout is never set if `go-plus.test.additionalTestFlags` is not set

lib/test/tester.js
```diff
@@ -222,6 +222,44 @@ class Tester {
...
 
+  buildGoTestArgs (timeoutMillis = 60000) {
+    const args = ['test', '-coverprofile=' + this.coverageFile]
+    const configFlags = atom.config.get('go-plus.test.additionalTestFlags')
+
+    let shortFlag = false
+    let verboseFlag = false
+
+    if (configFlags && configFlags.length) {
+      const arr = configFlags.trim().split(/\s+/g)
```
lloiser: parsing args this way is not correct, e.g.: `-my-arg="hello world"`

the [yargs-parser](https://github.com/yargs/yargs-parser) seems to parse them correctly

zmb3: Thanks for your review, and good catch. 👍

PR #569, opened 2017-02-17
fix tests for atom 1.15

lloiser:


joefitzgerald:
This looks like a really good idea 😉 .

PR #571, opened 2017-02-18
New feature: modify struct tags

zmb3:
## Description 
This PR adds support for @fatih's new `gomodifytags` tool.  
image

Two new commands are exposed: `golang:add-tags` and `golang:remove-tags`.  These can be used to automatically add/remove struct tags.  The UI offers options for specifying tags and their options, whether to use `snake_case` or `camelCase` for the tag names, and the ability to sort the tags.

A brief demo: (sorry about the super-speed gif)
gomodifytags

## Behavior

The behavior is similar to that of [vim-go](https://github.com/fatih/vim-go/pull/1204).

If the cursor is inside a struct and there is no selection, the tool operates on the entire struct.
If there is a selection then the tool only operates on line(s) that are touched by the selection (you don't need to select the entire line).

### Add Tags

If no tags are given then `json` is the default.  You can specify options such as `omitempty` in the options field, but this is not required.

### Remove Tags

If no tags are given then _all_ tags (and any corresponding options) are removed.
If tag names are given, then only those tags (and their options) will be removed.
If both tag and option are specified, then only the option is removed and the tag remains.

## TODO:

- [x] Add specs
- [x] Use `gomodifytags`'s JSON output and apply the updates to the buffer (instead of allowing the tool to rewrite the file for us - in some cases Atom doesn't refresh the buffer when this happens)
- [x] Test more complex scenarios with tag options and file any issues upstream.

### DIFF
--- a/lib/tags/gomodifytags.js
+++ b/lib/tags/gomodifytags.js
@@ -1,0 +1,175 @@
+'use babel'
+
+import {CompositeDisposable, Point} from 'atom'
+import {isValidEditor} from '../utils'
+import TagsDialog from './tags-dialog'
+
+export default class GoModifyTags {
+  constructor (goconfig) {
+    this.goconfig = goconfig
+    this.subscriptions = new CompositeDisposable()
+    this.subscriptions.add(atom.commands.add(
+      'atom-workspace', 'golang:add-tags',
+      () => this.commandInvoked('Add')
+    ))
+    this.subscriptions.add(atom.commands.add(
+      'atom-workspace', 'golang:remove-tags',
+      () => this.commandInvoked('Remove')
+    ))
+  }
+
+  dispose () {
+    this.subscriptions.dispose()
+    this.subscriptions = null
+    this.goconfig = null
+  }
+
+  commandInvoked (mode) {
+    const editor = atom.workspace.getActiveTextEditor()
+    if (!isValidEditor(editor)) {
+      return
+    }
+
+    if (editor.hasMultipleCursors()) {
+      atom.notifications.addWarning('go-plus', {
+        dismissable: true,
+        icon: 'tag',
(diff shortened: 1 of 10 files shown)


joefitzgerald:
This looks like a really helpful feature. Thanks @zmb3!!

zmb3:
@joefitzgerald @lloiser assuming CI completes (still queued) this is ready for review.

lib/tags/gomodifytags.js
```diff
@@ -0,0 +1,176 @@
...
+      archive += Buffer.byteLength(editor.getText(), 'utf8') + '\n'
+      archive += editor.getText()
+
+      executorOptions.input = archive
+    }
+
+    const args = this.buildArgs(editor, options, mode)
+
+    if (atom.inDevMode()) {
+      console.log('(go-plus): executing: gomodifytags ' + args.join(' '))
```
joefitzgerald: This statement seems to run while tests are executing, which pollutes CI output.

zmb3: We can remove it if you'd like.  I actually was thinking it would be nice if there was an option on the executor just to log the commands it executes.  I found it very helpful for troubleshooting when you want to run the commands outside of Atom and look at results.

Is there a similar check on the `atom` global that can tell me if I'm running specs?

joefitzgerald: > Is there a similar check on the atom global that can tell me if I'm running specs?

Nope, but we do set one (`go-plus.testing`) here: https://github.com/joefitzgerald/go-plus/blob/7c543f3b8cdc0636fd0369f4496a97699c2d8fcf/spec/spec-helpers.js#L23

joefitzgerald: I like the idea about having the executor print everything out in some kind of debug mode...

PR #578, opened 2017-02-25
:arrow-up: dependencies

zmb3:
Remove dependency on rimraf since fs-extra's `remove` function
performs the same duties.

Closes #572, #567, #566, #553, #552


zmb3:
1 of the 8 travis builds used an old version of node, hence the failure.  Is this something we can control?

joefitzgerald:
👍

PR #583, opened 2017-02-28
WIP: Add Guru Support

joefitzgerald:
This PR adds functionality based on guru:

- [x] Find usage, to show usage of identifiers in the go-plus panel
- [x] Highlight identifiers when the cursor changes
- [x] Ensure usage panel item initially displays progress when it is focused as a result of a find-usage command
- [x] Ensure that an error output is displayed in the usage panel when no identifier is found

Additional work could be done to expand the number of guru modes supported, or this could be shipped ahead of further expansion.

screen shot 2017-02-28 at 12 31 43 am

highlight-identifiers


lib/main.js
```diff
@@ -59,6 +60,7 @@ export default {
     this.gomodifytags = null
     this.godef = null
     this.hyperclickProvider = null
+    this.what = null
```
zmb3: `this.usage = null` too?

lib/panel/go-plus-panel.js
```diff
@@ -76,13 +77,19 @@ export default class GoPlusPanel extends EtchComponent {
         if (model && model.isActive) {
           model.isActive(true)
         }
+        if (model && model.setActive) {
+          model.setActive(true)
```
zmb3: What's the difference between `isActive` and `setActive`?

lib/usage/usage-view.js
```diff
@@ -0,0 +1,84 @@
...
+import EtchComponent from './../etch-component'
+import Octicon from 'etch-octicon'
+
+export default class UsageView extends EtchComponent {
+  constructor (props) {
+    super(props)
+    this.subscriptions = new CompositeDisposable()
+    if (props.model) {
+      props.model.view = this
+      props.model.updateContent('To find usage, select an identifier and run the `golang:find-usage` command via the command pallette.', 'initial')
```
zmb3: `props.model.updateContent(this.defaultContent(), 'initial')`

lib/usage/usage-view.js
```diff
@@ -0,0 +1,84 @@
...
+
+import {CompositeDisposable} from 'atom'
+import etch from 'etch'
+import EtchComponent from './../etch-component'
+import Octicon from 'etch-octicon'
+
+export default class UsageView extends EtchComponent {
+  constructor (props) {
+    super(props)
+    this.subscriptions = new CompositeDisposable()
```
zmb3: We're not subscribing to anything. Is this necessasry?

lib/what/what.js
```diff
@@ -0,0 +1,200 @@
...
+
+  unsubscribeToCursorEvents () {
+    this.clearMarkers()
+    if (this.cursorSubscriptions) {
+      this.cursorSubscriptions.dispose()
+      this.cursorSubscriptions = null
+    }
+  }
+
+  // parseStream (jsonStream) {
```
zmb3: Delete this?

lib/what/what.js
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  run (editor, pos, cursor, shouldDecorate) {
+    if (this.running) {
+      return
+    }
+
+    this.running = true
+    this.clearMarkers()
+    // TODO: clear markers
```
zmb3: TODO no longer necessary?

lib/what/what.js
joefitzgerald: Yes, it should probably be in guru-utils, and remove the logic about he index and instead return an array of stuff from index 0.

zmb3:
@joefitzgerald I think we can merge this now, assuming builds pass.

joefitzgerald:
Looks like we have a windows issue?

lib/guru-utils.js
```diff
@@ -0,0 +1,82 @@
...
+    return
+  }
+
+  const filePath = editor.getPath()
+  const args = ['-json']
+  if (scopedModes.includes(mode)) {
+    let relPath = atom.project.relativizePath(filePath)
+    if (relPath && relPath.length > 0 && relPath[0] !== null) {
+      // Make it relative to GOPATH because guru tool requires it.
+      if (this.goconfig && this.goconfig.locator) {
```
lloiser: `computeArgs` is a global function and it never receives a proper value for `this`. So this will never work.

lib/what/what.js
```diff
@@ -0,0 +1,142 @@
...
+    this.subscriptions = null
+    this.unsubscribeToCursorEvents()
+    this.goconfig = null
+    this.running = false
+  }
+
+  subscribeToCursorEvents () {
+    this.cursorSubscriptions = new CompositeDisposable()
+    this.cursorSubscriptions.add(atom.workspace.observeTextEditors((editor) => {
+      if (!editor || !editor.getBuffer()) {
```
lloiser: so this also runs for non-go editors?

styles/go-plus-panel.less
```diff
@@ -106,4 +106,16 @@
     justify-content: center;
     text-align: center;
   }
+
+  .go-plus-panel-body {
+    font-family: Menlo, Consolas, 'DejaVu Sans Mono', monospace;
```
lloiser: any reason why we don't/can't reuse atom's font settings (fontFamily, fontSize, lineHeight) ?

joefitzgerald: That's what we used to do. People complained that they wanted a different font size. The default font (if you don't set a fontFamily) is not monosoace.

lib/what/what.js
joefitzgerald: Good catch.

lloiser:
Another small issue with the highlighting of same variables: 
screen shot 2017-03-07 at 13 31 05
I've selected the code from back to front which then highlights `fmt`. So maybe do not highlight if there is a selection?

PR #594, opened 2017-03-06
added autocomplete-plus descriptions and URL

fasa123:
It finally works now: https://i.imgur.com/zR45JdV.png
I am calling atomParse which is this package here: https://github.com/fasa123/atomParse You can gladly implement that one directly into the go-plus package it rather should just be a quick fix. But it works very well. The URL thing also works. 

Only thing that could be done is executing atomParse asynchronously. Currently it can take couple seconds to load the docs for a huge package which really does not feel nice. But I could not figure out how to do that without blocking the UI and still waiting for the docs. Best bet would be to split it up so only load the docs for the currently selected autocomplete item if thats possible.

### DIFF
--- a/lib/autocomplete/gocodeprovider.js
+++ b/lib/autocomplete/gocodeprovider.js
@@ -237,12 +237,21 @@ class GocodeProvider {
     } catch (e) {
       console.log(e)
     }
+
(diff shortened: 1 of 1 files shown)


fasa123:
It still has some bugs (resolving a named declaration as a package). Maybe you can try to integrate this feature now though.

PR #597, opened 2017-03-08
WIP: Add Support For .go.json Configuration File

### LINKED ISSUE #76: Use different tool to "go build" per project
When working on golang projects that are meant to run in Google AppEngine (GAE), these projects must be compiled (and tested) with tools that ship with the SDK, namely `goapp`. This is a wrapper around the `go` binary, so it is almost fully compatible with `go` (i.e. you can run `goapp <CMD>` instead of `go <CMD>`, e.g. - `goapp test` instead of `go test`)

One way to work with this project in Atom (using go-plus) would be to add a symlink called "go" to point to the "goapp" binary, and set the installation path variable to the dir containing this symlink (thereby "fooling" the plugin to use the wrapper tool). However I believe this would persist the change for all projects (even non GAE ones).

Do you have suggestions on how I could make this change on a per-project basis? (Or maybe have this plugin detect GAE projects (using the index.yaml file in the top dir), and invoke the `goapp` tool instead (which is assumed to be in $PATH, or discoverable in the same way the go bin is detected).

### LINKED ISSUE #163: Support Strategy-Based Approach For Determining GOPATH In The Current Workspace
https://github.com/DisposaBoy/GoSublime/blob/master/USAGE.md#per-project-settings--project-based-gopath
It's useful if you want to put your projects outside of $GOPATH.

### LINKED ISSUE #207: Request: add support for gb
**gb** from @davecheney uses a vendoring approach which puts all the vendor code alongside the src in a project directory.

So it would be nice to be able to run `gb build` instead of `go build` so we can avoid all the `cannot find package...` warnings.

http://getgb.io/

(It would also be nice to be able to get rid of the `Warning: File [/Users/Andy/foo/src/foo/foo.go] does not reside within a "src" directory in your GOPATH...` warning somehow as well, but it's not quite as distracting.)

### LINKED ISSUE #281: Need a GoBuild cmd
I need a hot-key for go build my project, just build it no more else.
I hate build on save, i want control it my self.
And go1.5 is slower 2x than go1.4 on build.
It makes me crazy, thanks.

### LINKED ISSUE #308: Add default build / test flag support

### LINKED ISSUE #317: Checks errors outside of scope of project folder?
Not sure if this is a bug from gocode, go-plus, or user error. However I'm having the issue that when I save a source file and go-plus displays errors, it displays errors external to my project. 

For example, I have two projects in my GOPATH, all are Github clones and follow the language conventions for the directory layout. In my `src` folder I have this:
-  **Project A** is a personal project I am working on for class and sits in a private repo in `github.com/vivace-io/someproject`
-  **Project B** is (for example) an unbuilt clone of Gogs and sits in `github.com/gogits/gogs`

Project A is set as the project folder in my atom session and I am actively editing/saving these source files. Project B has been cloned but dependencies not downloaded with `go get`. So, when I save my changes in Project A, go-plus' debug window pops up and displays errors associated with not being able to find packages as well as any other code errors.

I'm using the latest releases of Atom's core packages, gocode, and go-plus (3.5.2) on a windows system. Removing and installing the go-plus package again resolves the issue, but only for the duration of the Atom session. I was also able to replicate this error on Ubuntu 15.10 as well. Obviously this can be resolved by running `go get` for those packages, and fix any errors. But when switching between projects it might become annoying. Any suggestions for possible fixes or what I might have done wrong?

Thanks!

### LINKED ISSUE #324: Option to Set GOPATH for each window/ session / project / file
Is it possible to have a `.goplus` file that GoPlus will search the directory hierarchy for and read GOPATH from it? 
I use different GOPATH (like env in python) for each project and changing GOPATH in the settings for each project is less than ideal.

### LINKED ISSUE #371: GAE Go App cannot find package
This plugin will continually report a package that cannot be found while working in an App Engine project within your GOPATH when the directory is underneath your application directory.

Example

goproject1
-- site.go
-- views
-- -- index.go

**site.go:**

```
package goproject1

import (
	"fmt"
	"net/http"
	"views"
)

func init() {
    http.HandleFunc("/", root)    
}

func root(writer http.ResponseWriter, request *http.Request) {
    fmt.Fprint(writer, "root")
    view := views.Index{Header: "My Header"} 
}
```

With views imported above the plugin throws an error:

**cannot find package "views" in any of: /usr/local/go/src/views (from $GOROOT) /Users/developer/Work/Projects/GoPlay/src/views (from $GOPATH)**

However the application compiles and runs as expected. So all in all everything is good from the code perspective except for the bogus errors being seen by the plugin.

On another note though, if I change the import as follows:

**site.go:**

```
package goproject1

import (
	"fmt"
	"net/http"
	"goproject1/views"
)

func init() {
    http.HandleFunc("/", root)    
}

func root(writer http.ResponseWriter, request *http.Request) {
    fmt.Fprint(writer, "root")
    view := views.Index{Header: "My Header"} 
}
```

The plugin lives happily, no errors, but the application does not compile. 

All in all I'll keep trucking with the errors from the plugin but makes knowing what is legit vs. not a little bit of a pain. Seems like it might just be an additional search path needed to be added such that its looking at GOROOT, GOPATH, and GOPATH/$PROJECT (just a thought).

### LINKED ISSUE #418: Windows with GOOS=linux fails to report bad syntax
I'm not sure where to continue my debugging.  Sorry if I'm in the wrong place/package/etc.

I'm used to running Atom in Linux, but ever since I got a 4k laptop, my usual QubesOS install requires a magnifying glass because apparently I can't change the resolution nor the DPI scaling for whatever reason.  So until that's resolved, I've resorted to using Windows 10 for a while, sadly.  And so I need to be able to write cross-platform code in my (temporary) Windows dev environment.

I'm used to just calling `GOOS=linux atom` or `GOOS=windows atom` when in Linux and as best as I can remember it worked well.  But when running in my current Windows 10 environment, when I call `set GOOS=windows; atom.exe` everything seems to work but `set GOOS=linux; atom.exe` doesn't.  The failure somewhere in the works where it decides whether or not code has valid syntax.  I figure that'd be `go build` or `go test` or `go fmt`.  When I save when using `GOOS=linux`, it just always returns "no issues" even when there are obviously issues.  I've supplied screenshots below to show the issue, which includes obviously bad code intended to trigger the error.

The reason I need to run `GOOS=..`. is because I want to ensure that the auto-complete and syntax checking are applicable to the target OS I'm writing for.  Usually I'm writing code for Linux, but sometimes I target Windows too.  And if I'm using a method that doesn't exist in my target OS, I'd sure like to know about it, and I'd also like to know about the methods that *are* available in my target, such as `syscall.NewLazyDLL` in Windows but not in Linux.

As a test, I've installed Quantum Shell (a package for Atom) that allows me to run commands in the Windows shell while in Atom.  When I type `set` it clearly shows the `GOOS` value that I expect to see.  When I run any usual commands, such as `go build`, it gives me the success or failure that I would have expected to see.  And because that's using the same environment variables that Atom is (supposedly) using, it seems that there's a bug somewhere, not just a bad environment variable on my part.

I'm running the following versions.  According to Atom, all packages are up-to-date.

Atom 1.7.1
go-plus 4.1.0
autocomplete-go 1.0.6
builder-go 1.0.0
go-config 1.1.4
go-get 1.0.3
gofmt 1.1.6
gometalinter-linter 1.1.0
gorename 1.0.2
linter 1.11.4
navigator-godef 1.0.2
quantum-shell 0.7.1
tester-go 1.0.4

I am running no other community packages whatsoever.

**Screenshots related to this issue:**

When in Windows launching Atom with `GOOS=linux`:
2016-04-15 03_18_12-test1 go c__users_dusty wilson_go-personal_src atom
(it should NOT show "no issues")

Same as above, but launching in Windows with `GOOS=windows`:
2016-04-15 03_19_27-test1 go c__users_dusty wilson_go-personal_src atom

When in Windows launching Atom with `GOOS=linux` (it doesn't appear when running `GOOS=windows`):
2016-04-15 03_06_09-test1 go c__users_dusty wilson_go-personal_src atom

After clicking the "Go Get" in the above alert:
2016-04-15 03_06_19-test1 go c__users_dusty wilson_go-personal_src atom

### LINKED ISSUE #551: Enhancement: Custom/Additional Test Flags (textbox)

### LINKED ISSUE #573: Does go-plus support build tags?
I've read the FAQ and searched issues, but can't find any info about it.

My build requires a build tag, e.g. `go build -tags 'production'` to compile successfully, is there a way to have go-plus do this for me?

At the moment I always have `1 issue` in the status bar with an error of `no buildable Go source files in ...`

Not a major problem if it's not supported, just wanted to make sure I haven't missed anything!

edit: p.s., thanks, go-plus is an incredibly incredibly useful plugin!

joefitzgerald:
🚨 WIP: DO NOT MERGE 🚨 

Adding support for a [`.go.json`](https://github.com/joefitzgerald/go-plus/wiki/.go.json-Project-Configuration-File-Format) configuration file will allow for a variety of scenarios where a user's default configuration in Atom differs from their requirements for a particular binary or library:

* Build tags
* Compilation for other `GOOS` values
* Specific environment variables that should be set
* Different arguments for `gometalinter`
* Etc...

This PR will support:

* Libraries (e.g. `github.com/joefitzgerald/somelibrary` with import path `github.com/joefitzgerald/somelibrary`)
* Package main (e.g. `github.com/joefitzgerald/somebinary` where the package is `main`)
* The `cmd/binary` convention (e.g. `github.com/joefitzgerald/somelibrary/cmd/somebinary` where the package is `main`)

The PR will require that you open atom projects at the repo level (e.g. `$GOPATH/src/github.com/joefitzgerald/somelibrary`).

Fixes #573 
Fixes #551 
Fixes #418 
Fixes #371 
Fixes #324 
Fixes #317 
Fixes #308 
Fixes #281 
Fixes #207 
Fixes #163 
Fixes #76

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,12 @@ You can add debug functionality to Atom by installing the following package:
 
 * [go-debug](https://atom.io/packages/go-debug): Debug your package / tests using [`delve`](https://github.com/derekparker/delve)
 
+## Per-Binary / Per-Library Configuration
+
+In some circumstances, you need to have a different configuration to the default
+for different binaries or libraries (e.g. to use build tags). In these
+circumstances you can use the `.go.json` file, which is documented here: https://github.com/joefitzgerald/go-plus/wiki/.go.json-Project-Configuration-File-Format.
+
 ## Builds
 
 ### How Are The Builds Performed?
--- a/lib/config/dotgodotjson.js
+++ b/lib/config/dotgodotjson.js
@@ -1,0 +1,49 @@
+'use babel'
+
+import fs from 'fs-extra'
+import path from 'path'
+
+class DotGoDotJson {
+  constructor (options) {
+    this.projects = new Map()
+  }
+
+  getProjectPaths () {
+    return atom.project.getPaths()
+  }
+
+  dispose () {
+    if (this.projects) {
+      this.projects.clear()
+    }
+    this.projects = null
+  }
+
+  search () {
+    const filesToCheck = []
+    for (const project of this.getProjectPaths()) {
+      filesToCheck.push(path.join(project, '.go.json'))
+      if (fs.existsSync(path.join(project, 'cmd'))) {
+        const cmds = fs.readdirSync(path.join(project, 'cmd')).filter(file => fs.statSync(path.join(project, 'cmd', file)).isDirectory())
+        for (const cmd of cmds) {
+          filesToCheck.push(path.join(project, 'cmd', cmd, '.go.json'))
+        }
+      }
+    }
+
+    const files = []
+    console.log(filesToCheck)
+    for (const file of filesToCheck) {
+      if (fs.existsSync(file)) {
+        files.push(file)
+      }
+    }
+
+    if (files && files.length) {
+      return files
+    }
+    return false
+  }
+}
+
+export {DotGoDotJson}
--- a/lib/config/settings.js
+++ b/lib/config/settings.js
@@ -3,7 +3,7 @@
 import fs from 'fs-extra'
 import path from 'path'
 
-class DotGoDotJson {
+class Settings {
   constructor (options) {
     this.projects = new Map()
   }
@@ -19,7 +19,7 @@ class DotGoDotJson {
     this.projects = null
   }
 
-  search () {
+  files () {
     const filesToCheck = []
     for (const project of this.getProjectPaths()) {
       filesToCheck.push(path.join(project, '.go.json'))
@@ -46,4 +46,4 @@ class DotGoDotJson {
   }
 }
 
-export {DotGoDotJson}
+export {Settings}
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "etch": "0.8.0",
     "etch-octicon": "^0.1.0",
     "fs-extra": "^2.0.0",
+    "fs-promise": "^2.0.0",
     "lodash": "^4.17.4",
     "semver": "^5.3.0",
     "temp": "^0.8.3"
--- a/spec/config/dotgodotjson-spec.js
+++ b/spec/config/dotgodotjson-spec.js
@@ -1,0 +1,67 @@
+'use babel'
+/* eslint-env jasmine */
+
+import {DotGoDotJson} from './../../lib/config/dotgodotjson'
+import {lifecycle} from './../spec-helpers'
+import path from 'path'
+import fs from 'fs-extra'
+
+describe('DotGoDotJson', () => {
+  let [dotgodotjson] = []
+  beforeEach(() => {
+    dotgodotjson = new DotGoDotJson()
+  })
+
+  afterEach(() => {
+    if (dotgodotjson) {
+      dotgodotjson.dispose()
+    }
+    dotgodotjson = null
+  })
+
+  describe('when there are no .go.json files in the project', () => {
+    describe('search', () => {
+      it('returns a falsy value', () => {
+        expect(dotgodotjson.search()).toBeFalsy()
+      })
+    })
+  })
+
+  describe('when there is one .go.json file in the root of a single project', () => {
+    let [project, filepath] = []
+    beforeEach(() => {
+      project = lifecycle.temp.mkdirSync('godotjson')
+      dotgodotjson.getProjectPaths = () => { return [project] }
+      filepath = path.join(project, '.go.json')
+      fs.writeFileSync(filepath, '{}')
+    })
+
+    describe('search', () => {
+      it('does not return a falsy value', () => {
+        const result = dotgodotjson.search()
+        expect(result).toBeTruthy()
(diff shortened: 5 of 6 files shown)


the1337guy:
Any updates?

joshua:
This would also be very useful for configuring a custom, per project GOPATH.

joshua:
@joefitzgerald Is this still a goal for go-plus?

I'm running into an issue trying to launch atom with a different GOPATH other then my "default".

`GOPATH=$(pwd) atom .` doens't appear to work as expected. gometalinter and goimports never run, for example.

Issue #550, opened 2017-01-14
Navigate to failed test


r03:
If a test fails, this is shown nicely in the test panel.
But is there an easy way to navigate to that failed test?
I was expecting that I could click on that error to jump to the correct file/position, but that doesn't work.

eduncan911:
The way I imagine this working (that is usable) is to parse the Test output for the error, and Hyperlink it to the file.

The problem with that analogy is not all failed test are parsable with the `go test` output.  For example, `t.Fail()` doesn't return anything:

```
$ go test -run TestFail
--- FAIL: TestFail (0.00s)
FAIL
exit status 1
```

Therefore, the only way this can work with that output is to parse the `functional name`, and some how search-and-link to it within the testing packages folders?  That's a lot of work.

Fatal()s are easier to find though:

```
$ go test -run TestFatal
--- FAIL: TestFatal (0.00s)
        podcast_test.go:70: it choked
FAIL
exit status 1
```

If file is already open in a tab, focus the tab and jump to line # in the test.

If file is not open, open in new tab, and jump to line # in the test.

But the problem is if using a testing package, like `testify/assert`.  It's output is vastly different depending on which assertion you are using:

```
$ go test -run TestFatal
--- FAIL: TestFatal (0.00s)
        Error Trace:    podcast_test.go:70
        Error:          Should be true
        Messages:       it choked
FAIL
```

But hey, there's that **Test Name** again.  

To me, this would mean the editor will need to keep track of all Test Names throughout all the open folders, in order to parse them for this single test name.  

This is possible with `grep`, using grep to -R recursive search; but, that would not be compatible with Windows machines.

r03:
I'm not a big vim user, but the test results seem to go to the quickfix windows where I can navigate to the failing test. (vim-go package) . 
Maybe we can use the same logic?

zmb3:
vim-go approach: https://github.com/fatih/vim-go/blob/0b6d16c572316deafbcf4139cd44757c9349c91d/autoload/go/tool.vim#L67

Seems to be just inspecting the output for anything that looks like a filename/line number, which suffers from the drawbacks mentioned above.

r03:
Maybe just make a hyperlink if a line matches something like this ? `(\w+_test\.go):(\d+)`
It doesn't cover everything, but probably the easiest to implement and covers most situations?

PR #599, opened 2017-03-12
Turn file references in the panel into links

zmb3:
Fixes: #550 

image

The test panel is now a legitimate etch component instead of raw HTML.


zmb3:
Investigate https://www.npmjs.com/package/react-ansi-style for restoring the color functionality.

lloiser:
fyi: I've tried to implement an atom grammar to display ANSI sequences in atom, see https://github.com/lloiser/language-ansi-styles. Maybe I can polish this a bit and use it here instead of ansi2html or react-ansi-style?

zmb3:
I'm not sure if a grammar would work.
That would mean the panel would have to be an editor wouldn't it?  If
that's the case then I'm not sure how to implement links.

zmb3:
ANSI functionality is restored and file links are working.
Need some specs and a way to make sure we don't get paths to temp files (from cover).

zmb3:
Submitted https://github.com/smallhelm/ansi-style-parser/pull/1 to add support for additional colors.  When this is merged we can add styles for the light versions. 

For example, with this PR and a new style, ginkgo's _light_ red output displays correctly:
```
.go-plus-ansi-lightRed {
   color: lighten(red, 20%);
}
```

zmb3:
This should be ready to go.  @lloiser you always seem to catch my mistakes after I merge, I'd love your review :-)

We still suffer from the filepaths sometimes being wrong due to coverage, but this only happens when there is a build error, not when a test fails (which was the original intent of #550).  I think we can handle that separately.

lib/test/test-panel.js
```diff
@@ -14,6 +19,36 @@ export default class TestPanel extends EtchComponent {
     if (this.props.model) {
       this.props.model.view = this
     }
+
+    this.locationRegex = /([\w-/.\\:]*.go:\d+)/g
```
lloiser: this does not capture a `<file>:<line>:<column>` like `foo/bar.go:23:4`
tbh: I haven't seen any output that contained this file syntax, so feel free to ignore this comment

spec/test/test-panel-spec.js
```diff
@@ -0,0 +1,71 @@
...
+import TestPanel from './../../lib/test/test-panel'
+
+describe('test panel', () => {
+  let testPanel
+
+  beforeEach(() => {
+    testPanel = new TestPanel()
+  })
+
+  describe('make link', () => {
```
lloiser: add two more tests:

_pre- and suffix_
input string: `"prefix foo/bar.go:23 suffix"`
output elements:
* SPAN with text `"prefix "`
* A with text `"foo/bar.go:23"`
* SPAN with text `" suffix"`

_multi line text_
input string: `"--- FAIL: TestFail (0.00s)\n\tbar_test.go:23: Error!\nFAIL"`
output elements:
* SPAN with text `"--- FAIL: TestFail (0.00s)\n\t"`
* A with text `"bar_test.go:23"`
* SPAN with text `": Error!\nFAIL"`

styles/ansi.less
```diff
@@ -0,0 +1,162 @@
...
+.go-plus-ansi-lightMagenta {
+  color: lighten(magenta, 20%);
+}
+
+.go-plus-ansi-lightCyan {
+  color: lighten(cyan, 20%);
+}
+
+.go-plus-ansi-lightWhite {
+  color: lighten(white, 20%);
```
lloiser: light white -> whiter than white 😆
maybe set normal white to a slightly greyish color and light white to white

lib/test/test-panel.js
zmb3: going to leave this as is - even with a column we'll still link to the correct line

lib/test/test-panel.js
```diff
@@ -27,10 +62,30 @@ export default class TestPanel extends EtchComponent {
...
-      <div ref='content' className='go-plus-test-panel' scrollTop={this.scrollHeight} style={style} innerHTML={output} />
+      <div ref='content' className='go-plus-test-panel' scrollTop={this.scrollHeight} style={style}>
+        <AnsiStyle text={output} mapText={this.makeLink.bind(this)} />
+      </div>
     )
   }
 
+  linkClicked (text, dir) {
+    const {file, line, column} = parseGoPosition(text)
+    const base = dir || projectPath()
```
zmb3: Good catch.  I did mean `this.props.dir`

lib/test/test-panel.js
```diff
@@ -27,10 +62,30 @@ export default class TestPanel extends EtchComponent {
...
+    if (!base) {
+      return
+    }
+
+    const filepath = path.join(base, file)
```
zmb3: :+1:

lib/test/test-panel.js
```diff
@@ -27,10 +62,30 @@ export default class TestPanel extends EtchComponent {
...
+    fs.access(filepath, fs.constants.F_OK | fs.constants.R_OK, (err) => {
```
zmb3: Agreed not necessary.

styles/ansi.less
zmb3: Good catch.  This was rather mechanical.

The white code is actually light gray, and "light white" is pure white.

zmb3:
I think everything's happy now, with the exception of:
- one of Lukas comments that I did not make code changes for 
- a flaky travis run (not sure why this happens but it seems to be a regular occurrence)

joefitzgerald:
You can ignore Travis failing on beta0 - it has an issue on linux that is known.

joefitzgerald:
This is awesome 👍 ⚡️ !

PR #609, opened 2017-03-28
Spring cleaning

zmb3:
Just a little cleanup.

See individual commits for details.

### DIFF
--- a/lib/build/builder.js
+++ b/lib/build/builder.js
@@ -1,6 +1,6 @@
 'use babel'
 
-import {CompositeDisposable} from 'atom'
+import {CompositeDisposable, Range} from 'atom'
 import fs from 'fs-extra'
 import path from 'path'
 import temp from 'temp'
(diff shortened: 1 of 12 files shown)


lib/build/builder.js
```diff
@@ -163,9 +159,9 @@ class Builder {
     const text = line[9]
     let range
     if (column && column >= 0) {
-      range = [[row - 1, column - 1], [row - 1, 1000]]
+      range = new Range([row - 1, column - 1], [row - 1, 1000])
```
zmb3: JSON.stringify is rather expensive. Serializing two range compatible arrays to strings and comparing the strings is a very heavy handed way of determining if they're equal. 

Range.isEqual amounts to a couple of integer comparisons. No allocations, no serialization. 

What about the cost of constructing the range objects you may ask?  We were paying for that anyway, just further down the chain. Eventually the range compatible array we pass to Atom gets turned into a range, were just doing that sooner.

PR #612, opened 2017-04-02
Handle save event in one place

### LINKED ISSUE #414: Don't run gometalinter if go build fails
When some code does not compile and if `gometalinter` is active, my computer gets very slow as all the linters are spitting tons of warnings and errors. It is also more difficult to spot the compilation errors among all the errors from the linters.

So I think `gometalinter` should only be run if go build succeed.

### LINKED ISSUE #484: tester-go ran tests multiple times
I had an issue where tester-go would run the tests 3 times at once. It took me a while to even realise this was happening, because it only shows a single run in the panel. 

I finally traced the problem so a single file, the _test.go I had been editing for a while. The problem occurred only when saving that file, but not when saving any other.

By chance I closed and reopened the file and the problem went away. 

If this should happen again, what could I do to help determine the cause of the bug and help fix it?

### LINKED ISSUE #526: Hide the test panel if code doesn't compile
Not sure if easily doable but it gets really annoying that the test panel is displayed when the code doesn't compile. I'm using the `Run Tests on Save` option.
Not that long ago, the panel wasn't displayed unless the code compiled. IMHO compile errors should only be displayed by the linter

zmb3:
This PR introduces a save event orchestrator that allows consumers to attach callbacks to two events:

- `onWillSave`
- `onDidSave`

`onWillSave` callbacks are invoked synchronously in the order they are registered, _prior_ to the editor being saved.  They return true to indicate success and allow the chain to continue, or return a non-true value to indicate an error that should prevent other save hooks from running.

The formatter is hooked up to `onWillSave`.

`onDidSave` callbacks are also invoked in the order they are registered, but these callbacks each return a promise.  If the promise resolves, future callbacks are invoked.  If the promise is rejected, the chain stops.

The build, test, and lint steps are hooked up to `onWillSave`

Note: as part of this PR, we are moving to linter's "indie" API, which allows us to push messages to linter when we are ready, instead of being asked for messages by linter.

Fixes #484
Fixes #414
Fixes #526


lib/autocomplete/gocodeprovider.js
```diff
@@ -61,7 +61,6 @@ class GocodeProvider {
     this.subscriptions = null
     this.goconfig = null
     this.subscribers = null
-    this.selector = null
```
joefitzgerald: ?

lib/build/builder.js
```diff
@@ -4,16 +4,14 @@ import {CompositeDisposable, Range} from 'atom'
 import fs from 'fs-extra'
 import path from 'path'
 import temp from 'temp'
+import {isValidEditor} from './../utils'
 
 class Builder {
-  constructor (goconfig) {
+  constructor (goconfig, linter) {
```
joefitzgerald: Is there a way to remove the linter dependency and replace it with some shim that lets us adjust approach in the future (e.g. diagnostics) in one place rather than many?

lib/build/builder.js
```diff
@@ -24,45 +22,36 @@ class Builder {
...
-        return Promise.all([buildPromise, testPromise]).then((results) => this.getMessages(results, options.cwd))
+        return Promise.all([buildPromise, testPromise])
       })
-    }).catch((error) => {
-      if (error.handle) {
-        error.handle()
+    }).then((results) => {
+      if (results && results.length) {
+        const messages = this.getMessages(results, options.cwd)
+        this.linter.setMessages(messages)
```
joefitzgerald: Don't we need to collect all messages and set them as a group? Otherwise only one stage can set messages?

lloiser: > Note: as part of this PR, we are moving to linter's "indie" API, which allows us to push messages to linter when we are ready, instead of being asked for messages by linter.

lib/autocomplete/gocodeprovider.js
zmb3: See the commit msg. The specs were printing a warning about our provider not being valid.

lib/build/builder.js
zmb3: Perhaps. This thing only has 3 functions so you could put a shim in there pretty easy.

lib/build/builder.js
zmb3: We are collecting all messages here, are we not?  Results is an array with an element for each stage.

joefitzgerald: @lloiser yes, I saw that. But either a) all messages from a "provider" are collected and you must explicitly clear them when they are no longer valid or b) you must collect all messages prior to setting them because each setMessages call implicitly clears all prior messages. In both cases I assume this means we can to centralize the setMessages call rather than invoking it multiple times per run.

zmb3: Set messages is only being invoked once here?

This is one thing v2 linter does better is distinguishes between set messages and set all messages.

joefitzgerald:
This is great, and I really only have these concerns:

1. We are conflating linter registration with orchestrator registration; these should be decoupled and deterministic
1. We are now showing up on timecop for load time because we've moved stuff outside of the bootstrap function and are requiring Orchestrator early
1. We are registering linters and registering with orchestrator when the linter is activated, which will prevent that functionality from working if linter is disabled
1. We currently register multiple linters and it may be better to just move the registration of the linter into the linter itself (e.g. gometalinter) so that it can be optional; alternatively, we may want to investigate moving to a single linter registration and aggregating all messages through a single component

spec/test/tester-spec.js
```diff
@@ -107,39 +107,6 @@ describe('tester', () => {
         expect(range.end.row).toBe(6)
         expect(range.end.column).toBe(1)
       })
-
```
joefitzgerald: Why the deletions here?

lloiser:
> 2. time cop

Move it into a `loadOrchestrator` function and call it whenever needed. It requires the script on demand and returns the cached orchestrator afterwards. (as it is done with all the other `load*` functions)

> 3. We are registering linters and registering with orchestrator when the linter is activated, which will prevent that functionality from working if linter is disabled

I think this ok, at least for the test linter and gometalinter. Building is still required because it enables autocompletion.

joefitzgerald:
We can't move it as written now because it is racy. We need to invert the way that linters register and defer the work to be done lazily later.

Tester has its own value regardless of whether linter is active or not. Additionally builder does too - it keeps autocomplete suggestions valid. It's only gometalinter that is truly optional.

zmb3:
Yes I agree we need to move registrations out of the linter callback.

Is there a hook you can add to your package that Atom runs after all
consumed services have been called?  Ideally that would be the place to do
the save registration.

spec/test/tester-spec.js
zmb3: This entire block was duplicated. Scroll up, the first copy is still there.

Issue #582, opened 2017-02-28
Improve UX for Update Tools command


zmb3:
The UX for a tools update could use some polish:
- the operation takes long enough that we should show some form of a progress update
- the notifications that come up when the operation completes take up a ton of space

I propose we focus the panel and print out each tool as it's being updated.  This will give the user a better indication that _something_ is happening, and we can do away with the notifications.

joefitzgerald:
:thumbsup:

PR #620, opened 2017-04-09
Create output panel

zmb3:
Convert the test panel into a general purpose output panel.

Connect build and test output to the output panel.

With this change the package is completely usable without linter, and it paves the way for displaying output for other events (for example, #582).

Here's a demonstration of the changes showing build and test failures with linter disabled.
Note that the ansi colors and clickable links from #599 are still supported.

output


zmb3:
Update tools integration:

update

lloiser:
Just a short review:
Can you please adjust the text for test lint (`go test -c`). I was really confused in the beginning that it "runs" the tests even though I've disabled it. Until I took a closer look at the actual text that says `-c` so it was the compilation/lint using `go test`.

zmb3:
Sure, how about `Compiling tests`?

lib/test/test-panel-manager.js
```diff
@@ -1,72 +0,0 @@
...
-    }
-    return this.emptyContent()
-  }
-
-  update (props) {
-    const oldProps = this.props
-    this.props = Object.assign({}, oldProps, props)
-
-    if (this.props.exitcode === 124) {
-      const timeoutMessage = 'Tests timed out after ' + atom.config.get('go-plus.test.timeout') + 'ms'
```
lloiser: This is gone but it was quite useful. Now an empty panel pops up if this happens.

lloiser:
👍  maybe also include the full command (in a smaller font-size?)

joefitzgerald:
Let's not have different file sizes....

Issue #630, opened 2017-05-02
'guru implements' returns 'guru failed' in the implements tab


veger:
### Description

The _guru_ command is installed by go-plus, I do not have it installed manually (as I do/did not use it)

When running _guru implements_, the developer console shows 
```
props running guru -json implements /home/user/go/src/etc/etc/my_file.go:#42

guru failed
```

Note that _golang:find-usages_ does work running
```
guru -json referrers /home/user/go/src/etc/etc/my_file.go:#42
```
as it properly shows the usages of the function I tried it on

### Output from `atom -v && apm -v`

Atom    : 1.16.0
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0
apm  1.16.1
npm  3.10.5
node 6.9.5 x64
python 2.7.12+
git 2.9.3

### Output From `go env`

GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/user/devel/go"
GORACE=""
GOROOT="/usr/lib/go-1.7"
GOTOOLDIR="/usr/lib/go-1.7/pkg/tool/linux_amd64"
CC="gcc"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build936633827=/tmp/go-build -gno-record-gcc-switches"
CXX="g++"
CGO_ENABLED="1"

### Steps to Reproduce

1. Run _Golang:implements_ command in a go file. In my case I tried it on a `type MyInterface interface` definition.

**Edit:** Hm... when running it on _MyInterface_ in the following line it actually does work:
```
func NewMyInterface(i ImplementationType) (MyInterface, error) {
```

Is this expected? (If so, the returned result _'guru failed'_ is confusing in this case...) Or is the _guru implements_ command partially broken?

zmb3:
The `-format` flag was changed to `-json` [a little over a year ago](https://github.com/golang/tools/commit/2da0720e4f0a8d82e510747c24a9e09038acb984).

It sounds like you just need to update, which you can do by either:
 - running `golang:update tools` in Atom
 - running `go get -u golang.org/x/tools/cmd/guru` in your shell

veger:
I just installed Atom and go-plus a couple of days ago. So I would expect that go-plus had installed the newest version of guru...

[...]

PR #633, opened 2017-05-03
Add 'time-out' message when a command timed out

veger:
This PR clarifies that a time-out occurred, solving confusion by the end user. See #630

### DIFF
--- a/lib/implements/implements.js
+++ b/lib/implements/implements.js
@@ -38,7 +38,7 @@ export default class Implements {
   }
 
   runGuru (args) {
-    const options = {timeout: 10000}
+    const options = {timeout: 20000}
     const archive = buildGuruArchive()
     if (archive && archive.length) {
       options.input = archive
--- a/lib/implements/implements.js
+++ b/lib/implements/implements.js
@@ -56,7 +56,11 @@ export default class Implements {
         return this.goconfig.executor.exec(cmd, args, options).then((r) => {
           if (r.error || r.exitcode !== 0 || (r.stderr && r.stderr.trim() !== '')) {
             if (this.view) {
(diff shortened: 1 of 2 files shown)


zmb3:
I think the executor should return stderr exactly as the process spit out, and we should detect the timeout downstream and display an appropriate message.

Are you up for making these changes?

veger:
Note that only the implements command checks (downstream) if the failing reason is a time-out. Other commands (check 'usage' only) does not (seem to) check this.

Should we update them all?

zmb3:
gorename and tester also check for timeouts.  You could also add a check to usage.

lib/implements/implements.js
```diff
@@ -55,8 +55,11 @@ export default class Implements {
         }
         return this.goconfig.executor.exec(cmd, args, options).then((r) => {
           if (r.error || r.exitcode !== 0 || (r.stderr && r.stderr.trim() !== '')) {
-            if (this.view) {
-              this.view.update('guru failed' + os.EOL + os.EOL + r.stderr.trim())
+              if (r.exitcode === 124) {
+                this.view.update('guru failed: time-out')
```
zmb3: How about:

```js
guru failed: operation timed out after `${operation.timeout} ms`
```

veger:
Updated 'implements' time-out with nicer message (and fixed position of return statement, found by linter :smile: ) and added similar implementation to 'usage'

zmb3:
Thanks!

tpjg:
Can this be added for more commands? E.g. when go-plus is doing it's lintTest ("go test -c -o {tempdir} .") it may also timeout, especially if CGO is involved.

Issue #644, opened 2017-05-10
Build timeouts are not shown


veger:
### Description

Time-out of (too) long builds are not shown in the output panel.

### Output from `atom -v && apm -v`

```
Atom    : 1.16.0
Electron: 1.4.16
Chrome  : 53.0.2785.143
Node    : 6.5.0
apm  1.18.1
npm  4.5.0
node 7.10.0 x64
python 2.7.13
git 2.12.2
```

### Steps to Reproduce

1. Fetch a large library (I used [go-gl/gl](https://github.com/go-gl/gl)), make sure it is not installed yet!
2. Use the library in your own project, so it will be build when your file is saved
3. Save and wait for the time-out

(or simply set the time-out config setting for the build command to a few ms :wink: )

### Expected Behavior

go-plus output shows the time-out message

### Actual Behavior

`Running go install .` disappears and output stays empty

PR #645, opened 2017-05-10
Show time-out message when build times out

veger:
Fix for #644 

1. `result.output` needs additional checking otherwise an empty string `""` results in `true`
2. If `result.output` is empty check if errorcode 124 is present, if so show time-out message

Note that the line (just after this fix)
```
return Promise.reject(new Error(result.output));
```
results in `builder.js [sm]:84 Uncaught (in promise) Error:` in the developer console.

I am no go-plus/promises expert, so I am unable to fix this... (so feel free to fix this in this PR if you actually know what is happening :wink: )

### DIFF
--- a/lib/build/builder.js
+++ b/lib/build/builder.js
@@ -69,11 +69,19 @@ class Builder {
           for (const result of results) {
             if (result.exitcode !== 0) {
               if (that.output) {
-                that.output.update({
-                  exitcode: result.exitcode,
-                  output: result.output,
-                  dir: options.cwd
-                })
+                if (result.exitcode === 124) {
+                  that.output.update({
+                    exitcode: result.exitcode,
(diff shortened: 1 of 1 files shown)


zmb3:
The timeout message we use everywhere is generally of the form:

_Operation timed out [after X ms]_.

Would you mind updating the wording here?

Don't worry about the error in the console, I have WIP that will clean that up.

veger:
Changed the output message

> Don't worry about the error in the console, I have WIP that will clean that up.

That is nice :+1:

veger:
Rebased PR, it is slightly different now since _lib/build/builder.js_ got some changes in master for which I needed to compensate

zmb3:
LGTM, thanks @veger!

PR #659, opened 2017-06-14
Asynchronous save in Atom 1.19+

zmb3:
Not sure when we want to merge this or if we need to make the changes in a more-backwards-compatible way.


lloiser:
Your changes should not cause any troubles for atom < 1.19
LGTM

PR #669, opened 2017-07-22
improve autocomplete provider

zmb3:
Don't merge this yet - would like to use it for a while and make sure everything works well.

Two changes here:
1. don't generate snippet tab stops for `...args` - these are often optional arguments and a tab stop disrupts the workflow.
2. support fuzzy matching, which has been requested numerous times over the past few years

image

For the fuzzy matching, we actually improve performance by calling gocode less.  We only invoke gocode when a `.` is entered or a new word begins, and we cache the results from that run.  For future keypresses, we refine the original set of suggestions using fuzzaldrin-plus.

### DIFF
--- a/lib/autocomplete/gocodeprovider.js
+++ b/lib/autocomplete/gocodeprovider.js
@@ -147,14 +147,14 @@ class GocodeProvider {
     let {prefix} = options
     prefix = prefix.trim()
     if (prefix === '') {
-      this.currentSuggestions = []
       if (!options.activatedManually) {
+        this.currentSuggestions = []
         const p = Promise.resolve([])
         this.notifySubscribers(p)
         return p
       }
     }
-    if (prefix.length > 1) {
+    if (prefix.length > 1 && this.currentSuggestions.length) {
       // fuzzy filter on this.currentSuggestions
       const p = new Promise((resolve) => {
         resolve(filter(this.currentSuggestions, prefix, {key: 'fuzzyMatch'})
--- a/lib/autocomplete/gocodeprovider.js
+++ b/lib/autocomplete/gocodeprovider.js
@@ -144,17 +144,24 @@ class GocodeProvider {
   getSuggestions (options) {
     // only invoke gocode when a new word starts or the '.' char is entered
     // on all other keystrokes we just fuzzy filter the previous set of suggestions
-    const {prefix} = options
-    if (prefix.trim() === '') {
+    let {prefix} = options
+    prefix = prefix.trim()
+    if (prefix === '') {
       this.currentSuggestions = []
-      return []
+      if (!options.activatedManually) {
+        const p = Promise.resolve([])
+        this.notifySubscribers(p)
+        return p
+      }
     }
-    if (prefix.length !== 1) {
+    if (prefix.length > 1) {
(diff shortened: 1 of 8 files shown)


lloiser:
It definitely feels faster and nicer with fuzzy matching!
Although I had some scenarios where
* I did not get the suggestions I expected 
* the popup for the suggestions did not open

I know these are vague descriptions but I have to narrow them down first 😞

zmb3:
Awesome. Would be great to figure out what these are. I'm holding off on
merging as I suspected there may be a few corner cases and I need time to
flesh it out.

lloiser:
Narrowed down one thing: 
given is this code:
```go
func helloWorld() {
    msg := "World"
    fmt.Printf(
        "Hello %s",
        // <-- cursor is right here in an empty line
    )
}
```
I often just press `ctrl+space` to give me all suggestions available here. Unfortunately this is not possible anymore because you already bail out early if the prefix is empty, which is the case for an empty line...

zmb3:
See if the latest commit helps you out @lloiser

zmb3:
@lloiser have you had a chance to use this a bit more?  Seems to be working well for me.

joefitzgerald:
@zmb3 mind if I rebase on master?

zmb3:
Not at all.

joefitzgerald:
The one enhancement I would like to make is: on the first keystroke after a period, we should run `gocode` without the next character to capture all possible completions, and then run it again with the character and merge the two result sets. This will allow the user to use any combination of letters to complete, even if they don't know the first letter of the completion they are shooting for.

Another way of thinking about it would be: we should always backtrack to the last period where one exists (without an intervening space) and run gocode from that point.

zmb3:
@joefitzgerald this should not be necessary.  When the `.` is pressed, we run gocode and get all valid suggestions.  From that point on we fuzzy filter until:

- manual invocation (ctrl+space): we always invoke gocode
- whitespace is entered: we clear the cached suggestions so a fresh gocode is run on the next keypress

VS Code has been using this approach for quite some time, so I'm pretty confident its sound.

joefitzgerald:
I think I see the issue. We are not providing suggestions for the first character after the dot in some scenarios. 

For example, try:
* Type `fmt.`
* Type `l`; you will only see autocomplete-snippets suggestions
* If you type `n` after the `l`, you will see the variety of `Scanln`, `Fscanln`, `Println`, etc. suggestions
* Type `fmt.` and then type `P` you will see all of the relevant `Print`, `Printf`, `Println` suggestions

zmb3:
Yeah, so one character is just not enough to do a fuzzy match I guess.

I don't think I mind not having every suggestion containing an 'l'
character to come up when I press the first 'l', but I admittedly haven't
tried it.

Maybe make the change in a branch and we can use it for a few days and see
if the experience is better.

zmb3:
Not able to reproduce that.
complete

joefitzgerald:
Turns out this is a difference in behavior for `gocode set autobuild false` (exhibits the behavior I was describing) vs. `gocode set autobuild true` (exhibits the behavior @zac captured in the gif above). This, despite `PostAuditLog` being present in the saved and installed code.

Issue #674, opened 2017-08-09
How to disable building?


Zamiell:
Thanks for the excellent plugin!

For my use-case, I want `go-plus` to run a format tool (`goimports`) on every save. But I don't want it to do any building, because my workflow uses the SFTP plugin - all the Go source files are on a remote server, and so building will always locally fail! So I'm constantly getting the panel popup flashing in my face every time I save a file.

How do I disable building??? There doesn't seem to be a check-box for this.
* I've unchecked "Run Tests On Save", and all of the other "Run Tests" related checkboxes, and that does not seem to stop building. I still get the panel popping up.
* Furthermore, I've added `--vendor, --disable-all, --json, .` to the gometalinter arguments box, and that doesn't seem to stop building either. I still get the panel popping up.

zmb3:
There's no way to do so currently, but this seems like a perfectly valid workflow.

I'll throw in a quick config option for this.

Zamiell:
Thank you for the quick response!!!!! <3

zmb3:
Created #675 to address this.  If you're comfortable giving this a try locally and can provide feedback that'd be great.

@joefitzgerald what do you think?  Make sense?

AlbinoGeek:
I need this, as otherwise `go-plus` always times out builds (after 60000) locking up my computer for up to a minute, for a file that builds manually in less than one second.  Something is broken with how `go-plus` calls the build.

zmb3:
@AlbinoGeek it definitely sounds like something's not right with your environment.  Are you able to share your screen sometime and show me what things look like for you?

Zamiell:
Thank you zmb3. Can you provide some short instructions on how to use this new feature?

zmb3:
@Zamiell this isn't yet released, so if you want to try it out you'll have to install go-plus from source.  If you can wait, we should have a new release out soon.

If not, start by uninstalling go-plus (either from the UI, or via `apm uninstall go-plus`.  Then clone this repo, `cd` into it, run `apm install`, followed by `apm link`.  Restart Atom and you should see a new setting "Disable Compilation on Save" that you can uncheck.  Note: you may also want to uncheck "run tests on save" because if your build fails then your tests probably won't run either.

joefitzgerald:
It’s acrjally much easier than that to work against master:

* `apm uninstall go-plus`
* `apm develop go-plus`
* `cd ~/github/go-plus`
* `apm link`

PR #675, opened 2017-08-10
add option to disable compilation on save

zmb3:
Add option `go-plus.config.compileOnSave` that defaults to true.

When disabled, go plus will not attempt to build your code when a file is saved.
In this mode, go-plus will only format the file on save, which can be helpful if for some reason local compilation would fail (see #674).

Additionally, we make the assumption that if the user doesn't want to compile on save, they also don't want to test on save (if local compilation would fail, local test runs would also fail), and so we never run tests on save if compile on save is disabled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -203,6 +203,13 @@
           "type": "string",
           "default": "",
           "order": 1
+        },
+        "compileOnSave": {
+          "title": "Compile on Save",
+          "description": "Compile the current package when a file is saved and provide linter errors.",
+          "type": "boolean",
+          "default": true,
+          "order": 2
         }
       }
     },
--- a/lib/build/builder.js
+++ b/lib/build/builder.js
@@ -42,6 +42,9 @@ class Builder {
   }
 
   build (editor, path = editor.getPath()) {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -203,6 +203,13 @@
           "type": "string",
           "default": "",
           "order": 1
+        },
+        "compileOnSave": {
+          "title": "Compile on Save",
+          "description": "Compile the current package when a file is saved and provide linter errors.",
```
joefitzgerald: You should clarify that if this is disabled `runTestsOnSave` will also have no effect. In general, I don't like conflating the two and think it would be more understandable for people to select each option independently.

zmb3:
Perhaps. I was afraid of people disabling this but leaving test on save
enabled and then running running into compiler errors.  Compilation is a
prerequisite for test so if you don't want to compile you cannot test.

xh3b4sd:
I want this. Do you plan to merge it?

Zamiell:
@zmb3 Is there a reason this isn't this merged yet?

PR #676, opened 2017-08-10
Show test coverage if go test’s exit code is non-zero

joefitzgerald:
The current behavior is to only show coverage when tests fail. This is an unnecessary restriction - we should allow coverage to be shown whenever the coverage file contains valid information.

Note: This does not allow coverage to be shown for a focused [ginkgo](https://github.com/onsi/ginkgo) spec, as the coverage file does not appear to be generated when any focusing occurs. See https://github.com/onsi/ginkgo/issues/205 for more details 😢.

### DIFF
--- a/lib/test/tester.js
+++ b/lib/test/tester.js
@@ -271,18 +271,17 @@ class Tester {
 
         output = output.trim()
         let state
-
         if (r.exitcode === 0) {
-          if (runTestsWithCoverage) {
-            this.ranges = parser.ranges(this.coverageFile)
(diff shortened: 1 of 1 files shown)


carldunham:
This seems to have reverted, for some reason. A quick trace through show it bailing at https://github.com/joefitzgerald/go-plus/blob/master/lib/test/tester.js#L125 with empty ranges.

carldunham:
Never mind. Looks like a ginkgo or go test issue. Sorry for the noise.

Issue #542, opened 2017-01-10
VSCode style tooltip


s4l1h:
Is it possible to do like vscode?

###  On MouseOver
![1](https://cloud.githubusercontent.com/assets/216654/21802678/c0d76d12-d737-11e6-8d2d-a385d2206742.png)
![2](https://cloud.githubusercontent.com/assets/216654/21802680/c0d81aaa-d737-11e6-8b78-00189a1a9b4c.png)
###  Ctrl+On MouseOver
![3](https://cloud.githubusercontent.com/assets/216654/21802679/c0d8018c-d737-11e6-8b62-85c32bf9b829.png)

fasa123:
At least the first thing is (kinda) possible. The autocomplete package has the fields suggestion.description and suggestion.descrptionMoreURL. I tried setting them in gocodeprovider.js.mapMessages but go doc requires the package name (or directory of it) as well as the function name to retrieve the documentation. I see no way of getting the package name (or its path) during the autocomplete method (at least none that I am able to implement with my terrible JS skills)

zmb3:
@fasa123 I looked at this and came to the same conclusion - it's not easy as `gocode` doesn't provide the package name.

There is an [issue](https://github.com/nsf/gocode/issues/415) to improve the suggestions, but whether or not it will be addressed is unclear to me.

fasa123:
Soo...
You can do something like this to actually fetch the documentation:

`    var pkg = editor.getTextInBufferRange([[position.row, 0], [position.row+1, position.column]]);
    pkg=pkg.substring(0, pkg.lastIndexOf(".")).replace(".", " ")
    console.log("Normalpkg: "+pkg)

      var cmd = 'go doc ' + pkg + ' ' + c.name;
      var doc = ''

      exec(cmd, function(error, stdout, stderr) {
        if(stdout.toString().indexOf(c.name)>0) {
          doc = stdout.toString()
        }
        console.log(doc)
      });
`

Though for some reason if I set suggestion.description = doc nothing shows. The console.log actually does show the documentation though but I cannot set it for the autocomplete provider. It just shows nothing.

s4l1h:
No more updates?

fasa123:
I can't do a pull request now but this works with every code I tested (gocodeprovider.js). This should not work with dot imports though. 


        var pkg = editor.getTextInBufferRange([[position.row, 0], [position.row + 1, position.column]]);
        var file = editor.getBuffer().file.path

[...]

PR #678, opened 2017-08-16
WIP: show descriptions in autocomplete suggestions

### LINKED ISSUE #351: Show documentation if available
Loop at this example in a java editor, when the autocomplete open, it shows too the related documentation to that attribute:

Image of Yaktocat

Is it possible to show the documentation of a method, if it has, in go-plus?

lloiser:
🚨 **WIP: do not merge** 🚨 

I've tried the [PR for gocode](https://github.com/nsf/gocode/pull/463) that adds a `package` property to it's suggestions.
This PR tries to add the actual description of the symbol to the autocomplete suggestion (requested in #542)

image

**Turns out it helps us, but it is definitely not the solution we want!**

I'm just leaving this PR here so that I've at least documented my findings somewhere :)

-----

**Problem 1**

We would still have to load all descriptions at once because autocomplete-plus does not support "lazy" loading of the description. Which puts quite some pressure on the computer (e.g. just trying to autocomplete `time.` takes several seconds to finish)
I even tried to outsmart them by using [property getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) for the `description` field, but they are accessing _all_ descriptions even before it is actually displayed, so no chance...

**Problem 2**

It is currently not possible to distinguish exported functions on structs and exported functions for the package if they have the same name. It's because we are only getting the name of the package but not the type of the struct we are currently getting suggestions for:
```go
package person

// Person represents a person
type Person struct{}

// Hello says hello person
func (ta Person) Hello() {}

// Hello says hello package
func Hello() {}
```

```go
package main

import "person"

func main() {
    person. // <- autocomplete here shows "says hello package"

    p := person.Person{}
    p. // <- also shows "says hello package" here!
}

```

**Problem 3**

Most of the standard go packages heavily use space indented comments to get some formatting into them.
Apparently this is not supported by autocomplete-plus yet.
We have to wait for https://github.com/atom/autocomplete-plus/commit/776f037dc214631d765dbdb8da56a561990e4ead to be shipped with atom.

Also once this has been added to atom, the description still sometimes wraps to a new line which is hard to read (especially for "formatted" documentation)

image

**Problem 4**

As you can see in the pictures above, the "signature" of the function (so `func Fprintf(....` is also shown in the description. Even though it would not be necessary because it is actually already visible in the rest of the suggestions anyway.
Unfortunately trying to get rid of the "signature" of the symbol we are asking `go doc` gets really complicated. The problem is that there are a variety of signatures that we would have to detect and get rid of.
e.g. Let's have a look at various go doc calls
```
$ go doc fmt Printf
func Printf(format string, a ...interface{}) (n int, err error)
    Printf formats according to a format specifier and writes to standard
    output. It returns the number of bytes written and any write error
    encountered.


$ go doc fmt Printf
type Stringer interface {
        String() string
}
    Stringer is implemented by any value that has a String method, which defines
    the ``native'' format for that value. The String method is used to print
    values passed as an operand to any format that accepts a string or to an
    unformatted printer such as Print.


$ go doc time UnixDate
const (
        ANSIC       = "Mon Jan _2 15:04:05 2006"
        UnixDate    = "Mon Jan _2 15:04:05 MST 2006"
        RubyDate    = "Mon Jan 02 15:04:05 -0700 2006"
        RFC822      = "02 Jan 06 15:04 MST"
        RFC822Z     = "02 Jan 06 15:04 -0700" // RFC822 with numeric zone
        // ... more consts here
)
    These are predefined layouts for use in Time.Format and Time.Parse. The
    reference time used in the layouts is the specific time:

    // ... more text here
```
I'm pretty sure I have not found all styles, but as you can see it could get pretty hard to parse all these different styles in order to get rid of the signature.

Fixes #351

### DIFF
--- a/lib/autocomplete/gocodeprovider.js
+++ b/lib/autocomplete/gocodeprovider.js
@@ -199,15 +199,17 @@ class GocodeProvider {
           if (r.stderr && r.stderr.trim() !== '') {
             console.log('autocomplete-go: (stderr) ' + r.stderr)
           }
-          let messages
+          let p = Promise.resolve()
           if (r.stdout && r.stdout.trim() !== '') {
-            messages = this.mapMessages(r.stdout, options.editor, options.bufferPosition)
+            p = this.mapMessages(r.stdout, options.editor, options.bufferPosition)
           }
-          if (!messages || messages.length < 1) {
-            return resolve()
-          }
-          this.currentSuggestions = messages
-          resolve(messages)
+          p.then((suggestions) => {
+            if (!suggestions || suggestions.length < 1) {
+              return resolve()
+            }
+            this.currentSuggestions = suggestions
+            resolve(suggestions)
+          })
         }).catch((e) => {
           console.log(e)
           resolve()
@@ -235,7 +237,7 @@ class GocodeProvider {
 
   mapMessages (data, editor, position) {
     if (!data) {
-      return []
+      return Promise.resolve([])
     }
     let res
     try {
@@ -249,13 +251,13 @@ class GocodeProvider {
         dismissable: true
       })
       console.log(e)
-      return []
+      return Promise.resolve([])
     }
 
     const numPrefix = res[0]
     const candidates = res[1]
     if (!candidates) {
-      return []
+      return Promise.resolve([])
     }
     if (candidates[0] && candidates[0].class === 'PANIC' && candidates[0].type === 'PANIC' && candidates[0].name === 'PANIC') {
       this.bounceGocode()
@@ -267,6 +269,7 @@ class GocodeProvider {
     } catch (e) {
       console.log(e)
     }
+    const promises = []
     const suggestions = []
     for (const c of candidates) {
       let suggestion = {
@@ -283,9 +286,12 @@ class GocodeProvider {
       if (suggestion.type === 'package') {
         suggestion.iconHTML = '<i class="icon-package"></i>'
       }
+      if (c.package) {
+        promises.push(this.getDescription(suggestion, c))
+      }
       suggestions.push(suggestion)
     }
-    return suggestions
+    return Promise.all(promises).then(() => suggestions)
   }
 
   translateType (type) {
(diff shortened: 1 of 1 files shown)


fasa123:
Just a random thought when I wrote https://github.com/joefitzgerald/go-plus/issues/542#issuecomment-305628262 I remember reading somewhere (I think it was a discussion on Microsoft/vscode-go) that one should use godef and godoc over "go doc". I cannot remember the reason and I cannot tell you whether this is still relevant I just wanted to highlight this.

zmb3:
Lots of great info @lloiser.  My thoughts:

I think problem 1 can be fixed with some updates to autocomplete-plus.  Joe knows a thing or two about that package.

Problem 2 seems like it should be an extension to the PR.  Perhaps `type` should differentiate between `func` and `method`.

Problem 4 can be solved with a little tooling instead of using `go doc` directly.  For example, in `gogetdoc` I return the decl and doc as separate fields:

```
~ ❯❯❯ gogetdoc -pos "/usr/local/Cellar/go/1.8.3/libexec/src/fmt/print.go:#5104" -json | jq
{
  "name": "Printf",
  "import": "fmt",
  "pkg": "fmt",
  "decl": "func Printf(format string, a ...interface{}) (n int, err error)",
  "doc": "Printf formats according to a format specifier and writes to standard output.\nIt returns the number of bytes written and any write error encountered.\n",
  "pos": "/usr/local/Cellar/go/1.8.3/libexec/src/fmt/print.go:189:6"
}
```

lloiser:
**UPDATE**

I have noticed a change in the autocomplete provider API:

`getSuggestionDetailsOnSelect`

> (experimental) Is called when a suggestion is selected by the user for the purpose of loading more information about the suggestion. Return a Promise of the new suggestion to replace it with or return null if no change is needed.

This sounds like a solution to problem 1.

Problem 3 has been resolved in the meanwhile too.

This leaves us only with 2 and 4 which are still open...

PR #685, opened 2017-08-30
autocomplete unimported packages

### LINKED ISSUE #665: Autocomplete on unimported package does not work
First of all, thank you so much for this great plugin. It really ease my migration to golang (I'm new to this language)

### Description

The "Autocomplete on unimported package" does not seem to work

### Output from `atom -v && apm -v`

Atom    : 1.18.0
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0
apm  1.18.1
npm  3.10.10
node 6.9.5 x64
python 2.7.12
git 2.7.4

### Output From `go env`

GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/my-user-name/go"
GORACE=""
GOROOT="/usr/local/go"
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build069586552=/tmp/go-build -gno-record-gcc-switches"
CXX="g++"
CGO_ENABLED="1"
PKG_CONFIG="pkg-config"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"

### Steps to Reproduce

1. Installed atom
2. Install go-plus package
3. Let it install dependency
4. create a go file, and start typing some unimported package name (ex : sqlx, nap, redigo.redis)

### Expected Behavior

autocomplete-plus shows the recomended package from gocode

### Actual Behavior

the autocomplete didn't trigger at all.. when i manually import packages, the autocomplete feature works flawlessly.

I remember using it without problem a few weeks ago when i first installed atom and go-plus, but now it seems that it doesn't work anymore.
I've tried using atom-beta, and do a clean install of both (only installing package go-plus and git-plus) but it still won't autocomplete unimported packages. I've also tried restarting gocode, reinstalling gocode (with deleting the bin and src folder) and deleting the gocode cache, it still won't work

Due to this, I'm forced to try to use vscode for the time being, but i really don't like the ui of that, still, the unimported autocomplete does work there which is strange, because they're using gocode too.

lloiser:
This is done by (temporarily) adding an `import "..."` statement into the code so that `gocode` can find the package and provide suggestions for it.

It is still rough on the edges especially for vendored packages.

Possible improvements:
* automatically add the `import "..."` when a suggestion is selected (nice to have, will be added by `goimports` on save anyway)
* the `pkg` map in `go.js` does not grow if new packages are added (low prio)

Fixes #665

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ install:
   - go get -u github.com/rogpeppe/godef
   - go get -u github.com/fatih/gomodifytags
   - go get -u golang.org/x/tools/cmd/guru
+  - go get -u github.com/tpng/gopkgs
 
 script:
   - 'curl -s https://raw.githubusercontent.com/atom/ci/master/build-package.sh | sh'
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -35,6 +35,7 @@ install:
   - go get -u github.com/rogpeppe/godef
   - go get -u github.com/fatih/gomodifytags
   - go get -u golang.org/x/tools/cmd/guru
+  - go get -u github.com/tpng/gopkgs
 
 build_script:
   - ps: iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/atom/ci/master/build-package.ps1'))
--- a/lib/autocomplete/gocodeprovider-helper.js
+++ b/lib/autocomplete/gocodeprovider-helper.js
@@ -1,0 +1,97 @@
+'use babel'
+
+import path from 'path'
+
+const vendorString = '/vendor/'
+
+export function wantedPackage (buffer, pos) {
+  // get the pkg the user tries to autocomplete from the current line
+  const lineTillPos = buffer.getTextInRange([[pos.row, 0], pos])
+  const matches = lineTillPos.match(/(\w+)\.$/)
+  if (!matches) {
+    return null
+  }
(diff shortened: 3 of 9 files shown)


zmb3:
Have you considered [`gopkgs`](https://github.com/tpng/gopkgs) instead of `go list`?  It reads `$GOPATH/pkg` instead of `$GOPATH/src` and is orders of magnitude faster.

lloiser:
Nice! I will check it out.

lloiser:
I choose [`gopkgs`](https://github.com/tpng/gopkgs) atm even though it does not resolve the actual package name but only the folder its in.
https://github.com/Microsoft/vscode-go/issues/647 shows that this can be an issue we will face sooner or later.

lloiser:
@zmb3 or @joefitzgerald can anybody pls review this 😁

joefitzgerald:
Thanks @lloiser!!

PR #699, opened 2017-10-26
allow test coverage for symlinks

everdev:
/[^/\\]+$/g finds the filename given a Windows or OS X path. We then
filter editorRanges to include those where the filenames on file and
r.file match, excluding _test.go files

### DIFF
--- a/lib/test/tester.js
+++ b/lib/test/tester.js
@@ -117,7 +117,10 @@ class Tester {
       return
     }
 
-    const editorRanges = _.filter(this.ranges, (r) => { return _.endsWith(file, r.file.replace(/^_\//g, '')) })
+		const re = /[^/\\]+$/g;
+		const editorRanges = _.filter(this.ranges, (r) => {
+		    return file.match(re)[0] == r.file.match(re)[0] && !_.endsWith(file, "_test.go")
+		})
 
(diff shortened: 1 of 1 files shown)


joefitzgerald:
Looks like you're using tabs on a spaces codebase... we use [`standard`](https://standardjs.com) to enforce style, which incidentally has a `--fix` flag that functions just like `gofmt`.

everdev:
ok, replaced tabs with spaces

lib/test/tester.js
```diff
@@ -117,7 +117,10 @@ class Tester {
       return
     }
 
-    const editorRanges = _.filter(this.ranges, (r) => { return _.endsWith(file, r.file.replace(/^_\//g, '')) })
+    const re = /[^/\\]+$/g
+    const editorRanges = _.filter(this.ranges, (r) => {
+      return file.match(re)[0] === r.file.match(re)[0] && !_.endsWith(file, '_test.go')
```
lloiser: Why is the check for _test.go necessary? It was not previously...

If it's really necessary then move it out of the `filter`.

everdev:
removed _test.go check as it's no longer necessary, good catch

joefitzgerald:
LGTM, thanks @everdev!

PR #702, opened 2017-10-31
Check for atom-ide in package manager, fixes #682

### LINKED ISSUE #682: Package recommendation displays when alt is installed
### Prerequisites

* [X] Have you tried launching `atom .` from the terminal in your project's directory?
* [X] Have you verified the output from `go env` is correct? If it is, please include the output in this issue.
* [X] Have you updated Atom to the latest version?
* [X] Have you tried using [Atom Beta](https://atom.io/beta), which can be run side-by-side with Atom Stable? Is the behavior the same, or different?
* [X] Have you updated your Atom packages to the latest versions?
* [X] Have you read the [FAQ](https://github.com/joefitzgerald/go-plus/wiki/FAQ)?
* [X] Have you searched [the issues](https://github.com/joefitzgerald/go-plus/issues?utf8=%E2%9C%93&q=is%3Aissue) to see if others are experiencing the same issue?

### Description

If you have `atom-ide-ui` installed you get:
> Additional features are available via the linter package. It runs linters and displays lint results. Alternatively, you can use Facebook's Nuclide package instead of the linter package.
screen shot 2017-08-23 at 10 17 41 am

You could choose "Never", but the description says `Alternatively, you can use Facebook's Nuclide package`. Since `atom-ide-ui` is installed, I would think that this shouldn't show up.

### Output from `atom -v && apm -v`

Atom    : 1.19.2
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0

apm  1.18.4
npm  3.10.10
node 6.9.5 x64
python 2.7.10
git 2.13.2

### Output From `go env`

```
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/lukescott/Documents/Go"
GORACE=""
GOROOT="/usr/local/Cellar/go/1.8.3/libexec"
GOTOOLDIR="/usr/local/Cellar/go/1.8.3/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/6h/thykxxkd70lgrv9vdc04sxt80000gn/T/go-build714148989=/tmp/go-build -gno-record-gcc-switches -fno-common"
CXX="clang++"
CGO_ENABLED="1"
PKG_CONFIG="pkg-config"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
```

### Steps to Reproduce

1. Disable `linter`, `linter-ui-default`, `intentions`, and `busy-signal`
2. Install `atom-ide-ui`
3. Restart Atom.

### Expected Behavior

go-plus should not recommend linter since the alternative is installed.

### Actual Behavior

go-plus recommends `linter` or `atom-ide-ui` even though `atom-ide-ui` is installed.

(Also wondering if it should make sure both are not enabled at the same time)

lukescott:
Fixes #682. Added a bundledAlt map to check for alternative packages. Also changed verbiage of linter message since atom-ide-ui is maintained by Atom now.

### DIFF
--- a/lib/package-manager.js
+++ b/lib/package-manager.js
@@ -21,7 +21,11 @@ const bundledPackages = new Map([
   ['hyperclick', 'It enables alt-click for go to definition, and shift-alt-click to return to the prior location.'],
   ['go-debug', 'It allows you to interactively debug your go program and tests using delve.'],
   ['go-signature-statusbar', 'It shows function signature information in the status bar.'],
-  ['linter', 'It runs linters and displays lint results. Alternatively, you can use Facebook\'s Nuclide package instead of the linter package.']
+  ['linter', 'It runs linters and displays lint results. Alternatively, you can use atom-ide-ui.']
+])
+
+const bundledAlt = new Map([
+  ['linter', 'atom-ide-ui']
 ])
 
 const goTools = new Map([
@@ -102,9 +106,17 @@ class PackageManager {
         continue
       }
 
+      const altPkg = bundledAlt.get(pkg)
+      if (altPkg) {
+        const ap = atom.packages.getLoadedPackage(altPkg)
+        if (ap) {
+          continue
+        }
+      }
+
       let disabled = false
       for (const d of atom.config.get('go-plus.disabledBundledPackages')) {
-        if (d && d.trim() === pkg) {
+        if (d && [pkg, altPkg].includes(d.trim())) {
           disabled = true
           break
         }
--- a/lib/package-manager.js
+++ b/lib/package-manager.js
@@ -21,11 +21,11 @@ const bundledPackages = new Map([
   ['hyperclick', 'It enables alt-click for go to definition, and shift-alt-click to return to the prior location.'],
   ['go-debug', 'It allows you to interactively debug your go program and tests using delve.'],
   ['go-signature-statusbar', 'It shows function signature information in the status bar.'],
-  ['linter', 'It runs linters and displays lint results. Alternatively, you can use atom-ide-ui.']
+  ['atom-ide-ui', 'It provides language ide features.']
 ])
 
-const bundledAlt = new Map([
-  ['linter', 'atom-ide-ui']
+const altPackages = new Map([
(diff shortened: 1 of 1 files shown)


lib/package-manager.js
```diff
@@ -21,7 +21,11 @@ const bundledPackages = new Map([
   ['hyperclick', 'It enables alt-click for go to definition, and shift-alt-click to return to the prior location.'],
   ['go-debug', 'It allows you to interactively debug your go program and tests using delve.'],
   ['go-signature-statusbar', 'It shows function signature information in the status bar.'],
-  ['linter', 'It runs linters and displays lint results. Alternatively, you can use Facebook\'s Nuclide package instead of the linter package.']
+  ['linter', 'It runs linters and displays lint results. Alternatively, you can use atom-ide-ui.']
```
joefitzgerald: I would prefer to replace `linter` in this list with `atom-ide-ui`. Does that simplify this somewhat?

lukescott:
I could make atom-ide-ui the primary. Should allow linter to be an alternate though, at least for now. You wouldn’t want the message to display for those using linter instead.

joefitzgerald:
I think I'm OK displaying it for people using linter. Most (historical) users of atom already have linter installed, and we will be adding more features to go-plus that require `atom-ide-ui`, so this seems reasonable. The user is welcome to say "never" when prompted to install, and then they will never see it again.

lukescott:
Last commit defaults to `atom-ide-ui` and displays a warning if `linter` is installed. I could simply change it to `atom-ide-ui`, but thought it might not be a good idea to install `atom-ide-ui` on top of an existing `linter`, thus the warning (wasn't sure what text to put there).

PR #716, opened 2017-12-05
WIP: fix some invalid suggestions

lloiser:
🚨  **WIP** 🚨 

This merge request attempts to fix some invalid suggestions that exist since the switch to fuzzy matching and caching of `gocode` results.

I will add more comments on how to reproduce a certain problem and (if easily possible) also a unittest (that obviously will fail for the moment)

### DIFF
--- a/lib/autocomplete/gocodeprovider.js
+++ b/lib/autocomplete/gocodeprovider.js
@@ -198,9 +198,20 @@ class GocodeProvider implements AutocompleteProvider {
     const sameFile = this.currentFile === editor.getPath()
     const sameLine = this.currentRow >= 0 && this.currentRow === bufferPosition.row
     const movingForward = this.currentColumn >= 0 && this.currentColumn <= bufferPosition.column
-    const useCache = !options.activatedManually && sameFile && sameLine && movingForward
+    let useCache = !options.activatedManually && sameFile && sameLine && movingForward
 
-    if (useCache && this.currentSuggestions.length && prefix.length > 0 && !prefix.endsWith('.')) {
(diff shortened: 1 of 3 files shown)


lloiser:
### GIVEN

```go
import "fmt"

func main() {
	foo := "bar"
	fmt.Printf(foo)

}
```

The cursor is currently at the empty line after `fmt.Printf`

### WHEN

I start typing `fmt.Printf(` (character by character)

And finally write `f` again.

### THEN

**Expected**

I get a suggestion for `foo`

**Actual**

I get (again) suggestions for the `fmt` package like `Printf`, `Errorf`, etc...

joefitzgerald:
What else needs to be done on this?

lloiser:
I still get wrong suggestions from time to time but it's really difficult to reproduce it and even harder in a unittest...

tbh I think we should change the current behaviour to match the one implemented in the atom-languageclient: https://github.com/atom/atom-languageclient/pull/128. It seems like a sound solution.

lloiser:
### GIVEN

```go
import "fmt"

func main() {
	fmt.Errorf("my error")
}
```

The cursor is currently at the empty line after `fmt.Printf`

### WHEN

add a `.` after `fmt.Errorf("asdf")`

### THEN

**Expected**

I get a suggestions for the returned error, so `String()`

**Actual**

I get nothing

tj:
I can't quite narrow down a specific pattern, but matching definitely seems a bit messed up lately, even if you have a `Decoder` struct in the file for example it'll just ignore it entirely (unless that's a gocode issue)

Issue #681, opened 2017-08-23
Package add dialog


lukescott:
### Prerequisites

* [X] Have you tried launching `atom .` from the terminal in your project's directory?
* [X] Have you verified the output from `go env` is correct? If it is, please include the output in this issue.
* [X] Have you updated Atom to the latest version?
* [X] Have you tried using [Atom Beta](https://atom.io/beta), which can be run side-by-side with Atom Stable? Is the behavior the same, or different?
* [X] Have you updated your Atom packages to the latest versions?
* [X] Have you read the [FAQ](https://github.com/joefitzgerald/go-plus/wiki/FAQ)?
* [X] Have you searched [the issues](https://github.com/joefitzgerald/go-plus/issues?utf8=%E2%9C%93&q=is%3Aissue) to see if others are experiencing the same issue?

### Description

Feature: Add a dialog menu similar to GoSublime's Add/Remove package:

<img width="752" alt="screen shot 2017-08-23 at 9 55 34 am" src="https://user-images.githubusercontent.com/486224/29628167-8a6c0b80-87e9-11e7-97eb-e2258430d4eb.png">

In GoSublime you can activate this two different ways:

- `Cmd-Shift-P`, type add package, Enter, then choose from dialog
- `Cmd-. + P`, then choose from dialog

### Use Case

goimports adds/removes packages as you save, but this user experience can sometimes be frustrating.

Use Case 1) 

When two packages share symbol names. For example, I have a package within my repo called "backend" and put "backend.Backend" in a struct. goimports decides to import etcd's backend package instead, forcing me to manually resolve it.

If I remove backend.Backend temporarily and save, it removes the import. When I add it back and save again it goes back to etcd's import.

This could be improved by:

1 - Looking in the same package at other files to see if the same package name was imported. Resolve conflicts by using the one already in the file.

2 - If an import was used in a previous save and is later re-added, remember it and use it again. If I am using the package for the first time there may be an issue there - but at least it would remember my choice after I manually resolved it the first time.

Why would I remove a package? Because Go forces me to when I remove an identifier.

The possible conflict on the second point would be resolved with the dialog feature. I could remove and add the package very quickly without having to manually change the path.

Use Case 2)

When you type the `package.Func`, then Cmd-Save auto-complete doesn't work. You have to delete Func and re-type it for it to find it. This is because the package was not imported when you first typed `package.Func`.

[...]

Issue #656, opened 2017-05-28
Add Package Import Palette


leonklingele:
Please add support for a `Package Import Palette` as implemented in [GoSublime](https://github.com/DisposaBoy/GoSublime).

When pressing `super+. super+p`, GoSublime shows a searchable modal with every installed Go package in it. This makes it easy to find packages to import.
It would be awesome to have such a palette in go-plus as well.

See this screenshot (pressing _Enter_ will add the package to the import list):
![gosublime-package-import-palette](https://i.imgur.com/wWtjkkx.png)

Issue #161, opened 2015-02-26
List all available packages


carlosbrando:
GoSublime has this list that you can use to add and delete packages. I see this as an important addition to this awesome package as well.

![screen shot 2015-02-26 at 14 04 02](https://cloud.githubusercontent.com/assets/2035/6396981/93166c26-bdc0-11e4-9146-bd74d513ca49.png)

joefitzgerald:
`goimports` largely obviates the need for this...

carlosbrando:
Yeah, but to `goimports` do its job it's necessary to remember the name of the package and the exact name of the function you are looking for without using the autocomplete function.

Is it possible to autocomplete suggest possible imports when autocomplete fails to find a match?

joefitzgerald:
This sounds like a great use case for an autocomplete-plus provider, that can help you with defining packages. It would only return suggestions when inside an import block. We might need to adjust the grammar to explicitly define when you are importing. Right now I see the cursor having the following scope inside an punctuation.definition.string.end.goimport("...")` or `import ""` block: `punctuation.definition.string.end.go`, so that would be an easy thing to check initially.

firstrow:
autocomplete imports - must have.

marconi:
+1 to this, I miss this coming from `GoSublime`

fmpwizard:
another +1 for a similar functionality to goSublime, I had to disable `goimport` because here and there it would lock up Atom (On fedora 20, 64bit , atom 1.0.2 go-plus 3.4.1), so now I have gofmt to format my files, and manually add imports. Other than this, go-plus is awesome!

geovanisouza92:
+1

mastercactapus:
+1 the most useful part IMO is being able to add/remove imports without losing my place in the current file, it's hard to go back to that :)

VitalySenko:
+1 That would be cool!

arlimus:
+1, the auto-completion of imports from goimports doesn't always get it right. when it fails, the goimports-on-save becomes a real annoyance ;) go sublime was great with the package selection feature

zmb3:
Closing as this functionality is covered in #226

PR #720, opened 2017-12-09
Add fuzzy import dialog

### LINKED ISSUE #718: Fuzzy import search
### Prerequisites

* [x] Have you tried launching `atom .` from the terminal in your project's directory?
* [x] Have you verified the output from `go env` is correct? If it is, please include the output in this issue.
* [x] Have you updated Atom to the latest version?
* [x] Have you tried using [Atom Beta](https://atom.io/beta), which can be run side-by-side with Atom Stable? Is the behavior the same, or different?
* [x] Have you updated your Atom packages to the latest versions?
* [x] Have you read the [FAQ](https://github.com/joefitzgerald/go-plus/wiki/FAQ)?
* [x] Have you searched [the issues](https://github.com/joefitzgerald/go-plus/issues?utf8=%E2%9C%93&q=is%3Aissue) to see if others are experiencing the same issue?

### Description

Fuzzy search for imports, primarily because it's difficult to remember usernames.

### Expected Behavior

Typing "uuid" for example would show all uuid packages locally. It would be potentially even cooler to list godoc.org's as well.

### Actual Behavior

Nothing

---

Just an idea :D

zmb3:
Present a dialog that fuzzy matches packages available to import (from `$GOPATH` or the project's vendor directory).

goplusimport

Fixes #718, #681, #656, and #161.


tj:
niceeeeee 🍺

lib/simple-dialog.js
```diff
@@ -0,0 +1,83 @@
...
+import EtchComponent from './etch-component'
+
+type Props = {
+  prompt: string,
+  initialValue?: string,
+  onConfirm: (string) => void,
+  onCancel?: () => void
+}
+
+export default class SimpleDialog extends EtchComponent {
```
lloiser: Maybe we want to change this to actually have two components: `Dialog`, and `TextEditorDialog`

The `Dialog` component just displays any kind of children (passed in via the constructor) inside the `go-plus-dialog` div (inside an atom modal of course)

The `TextInputDialog` just uses the `Dialog` and passes the div with the prompt and the TextEditor as children to `Dialog`.

`Dialog` then could for example also be used for the `tags` dialog.

This is just a suggestion. I haven't thought it through yet. It could also be done later once we find the time and we need it 😉

lib/simple-dialog.js
```diff
@@ -0,0 +1,83 @@
...
+  props: Props
+  subscriptions: CompositeDisposable
+  panel: any
+  element: any
```
lloiser: This should be moved to `EtchComponent` and it is actually an `HTMLElement`

spec/import/importer-spec.js
```diff
@@ -0,0 +1,88 @@
...
+})
+
+describe('importer', () => {
+  let importer = null
+  let editor = null
+
+  beforeEach(() => {
+    runs(() => {
+      lifecycle.setup()
+      console.log('lifecycle setup complete')
```
lloiser: unnecessary

lib/import/importer.js
```diff
@@ -0,0 +1,119 @@
...
+import type {GoConfig} from './../config/service'
+
+// Filters an array of all possible import paths into those that are importable
+// from some source package.
+const importablePackages = (sourceImportPath: string, packages: Array<string>): Array<string> => {
+  return packages.filter(pkg => {
+    // filter out unimportable vendor and internal packages
+    // https://golang.org/cmd/go/#hdr-Internal_Directories
+    // https://golang.org/cmd/go/#hdr-Vendor_Directories
+    const vendor = pkg.indexOf('/vendor/')
```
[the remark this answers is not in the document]
zmb3: Great, I couldn't really grok what you had done and how similar this was.

lib/import/importer.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  dispose () {
+    this.subscriptions.dispose()
+    this.subscriptions = null
+  }
+
+  commandInvoked () {
+    const pkgMap: Map<string, Array<string>> = allPackages(this.goconfig)
+    const pkgs = [].concat.apply([], Array.from(pkgMap.values()))
```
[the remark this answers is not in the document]
zmb3: `Array.from(pkgMap.values())` gives you an array of arrays.  We need to flatten it into a single array.

lib/import/importer.js
```diff
@@ -0,0 +1,119 @@
...
+    if (r.error) {
+      atom.notifications.addError('Error', {
+        detail: r.error.message,
+        dismissable: true
+      })
+      return {success: false, r}
+    }
+
+    if (r.exitcode === 0) {
+      editor.getBuffer().setTextViaDiff(r.stdout)
```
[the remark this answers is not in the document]
zmb3: Yeah I would ultimately like to make it just string, but why does that matter here?  There's no type error is there?

lib/simple-dialog.js
zmb3: Seems overkill for now.  In fact I really didn't even need to make simple dialog as part of this PR.  When I first started out I thought I was going to need to make my own view for the drop down but then remembered `atom-select-list`.

scbizu:
Hi there, I found that  if there is a error in reading directory(like `permission denied`) ,the whole fuzzy finder will crash.

PR #735, opened 2018-01-15
add config options for addditional build/test args

### LINKED ISSUE #282: Cover doesn't work always
I'm encountering an issue where cover coloring doesn't work always. In small packages it works, no problems. But in 2 cases no color shows up:
- Package that builds slowly because it's using C code using cgo.
- Package that is has a bunch of imports, also including the package which uses cgo, I just mentioned.

There are no compile errors and all tests pass.

Can anyone give some pointers to investigate the issue?

**UPDATE**
It seems it has to do with the time it takes to execute the tests. Is it an idea to run by default the tests with `-short`? Or make it a setting?
## 

go-plus information:

```
Go: go1.5.1 darwin/amd64 (@/usr/local/go/bin/go)
GOPATH: /path/to/gopath
Cover Tool: /usr/local/go/pkg/tool/darwin_amd64/cover
Vet Tool: /usr/local/go/pkg/tool/darwin_amd64/vet
Format Tool: /path/to/gopath/bin/goimports
Lint Tool: /path/to/gopath/bin/golint
Gocode Tool: /path/to/gopath/bin/gocode
Godef Tool: /path/to/gopath/bin/godef
Gocode Status: Enabled
Oracle Tool: /path/to/gopath/bin/oracle
Git: /usr/bin/git
PATH: /usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/usr/local/go/bin
Atom: 1.0.14 (darwin x64 14.5.0)
```

### LINKED ISSUE #308: Add default build / test flag support

### LINKED ISSUE #551: Enhancement: Custom/Additional Test Flags (textbox)

### LINKED ISSUE #573: Does go-plus support build tags?

### LINKED ISSUE #690: Run 'go test' with -race flag
### Description

In the GUI settings I'm unable to set ```go-test``` to run with ```-race``` on save. Is this by design ?

zmb3:
Introduce `go-plus.config.additional{Build|Test}Args` config properties that allow users to alter the flags that are passed to:

build args:
- go build / go install
- go test -c

test args:
- go test -coverprofile

This can be used to define build tags, for example.

- Fixes #282
- Fixes #308
- Fixes #551
- Fixes #573
- Fixes #690
- Closes #565

### DIFF
--- a/lib/build/builder.js
+++ b/lib/build/builder.js
@@ -1,10 +1,11 @@
 // @flow
 
-import {CompositeDisposable, Range} from 'atom'
+import argparser from 'yargs-parser/lib/tokenize-arg-string'
 import fs from 'fs-extra'
+import os from 'os'
 import path from 'path'
 import temp from 'temp'
-import os from 'os'
+import {CompositeDisposable, Range} from 'atom'
 import {isValidEditor} from './../utils'
 import {getgopath} from '../config/environment'
 
@@ -152,6 +153,12 @@ class Builder {
       buildArgs.push('-o')
       buildArgs.push(this.devNull())
     }
+    const additionalArgs = atom.config.get('go-plus.config.additionalBuildArgs')
+    if (additionalArgs && additionalArgs.length) {
+      const parsed = argparser(additionalArgs)
+      buildArgs.push(...parsed)
+    }
+
     buildArgs.push('.')
     this.output.update({
       output: 'Running go ' + buildArgs.join(' '),
@@ -177,8 +184,29 @@ class Builder {
     return process.platform === 'win32' ? 'NUL' : '/dev/null'
   }
 
+  testCompileArgs (additionalArgs: string = ''): Array<string> {
+    const result = ['test']
+    // use additional build args even when we compile the tests
+    if (additionalArgs && additionalArgs.length) {
+      const parsed = argparser(additionalArgs)
+      for (let i = 0; i < parsed.length; i++) {
+        if (parsed[i] === '-o') {
+          // we'll take care of this one, skip over the -o flag
+          i++
+          continue
+        } else if (parsed[i] === '-c' || parsed[i] === '-i') {
+          continue
+        } else {
+          result.push(parsed[i])
+        }
+      }
+    }
+    result.push('-c', '-i', '-o', this.devNull(), '.')
+    return result
+  }
+
   async lintTest (cmd: string, options: ExecutorOptions) {
-    const testArgs = ['test', '-c', '-i', '-o', this.devNull(), '.']
+    const testArgs = this.testCompileArgs(atom.config.get('go-plus.config.additionalBuildArgs'))
 
     this.output.update({
       output: 'Compiling tests:' + os.EOL + '$ go ' + testArgs.join(' '),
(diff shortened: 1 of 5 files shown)


lloiser:
The code looks good :+1:
I will test it soon!

zmb3:
Planning on merging this tomorrow so let me know if anyone spots issues.

lib/build/builder.js
```diff
@@ -177,8 +184,29 @@ class Builder {
     return process.platform === 'win32' ? 'NUL' : '/dev/null'
   }
 
+  testCompileArgs (additionalArgs: string = ''): Array<string> {
+    const result = ['test']
+    // use additional build args even when we compile the tests
+    if (additionalArgs && additionalArgs.length) {
+      const parsed = argparser(additionalArgs)
+      for (let i = 0; i < parsed.length; i++) {
```
joefitzgerald: `for...of`?

lib/build/builder.js
```diff
@@ -177,8 +184,29 @@ class Builder {
...
+        if (parsed[i] === '-o') {
```
joefitzgerald: Should we trim?

lib/build/builder.js
```diff
@@ -177,8 +184,29 @@ class Builder {
...
+          // we'll take care of this one, skip over the -o flag
+          i++
+          continue
+        } else if (parsed[i] === '-c' || parsed[i] === '-i') {
```
joefitzgerald: Trim?

lib/test/tester.js
```diff
@@ -211,6 +212,45 @@ class Tester {
     this.coverageFile = path.join(this.tempDir, 'coverage.out')
   }
 
+  buildGoTestArgs (timeoutMillis: number = 60000): Array<string> {
+    const args = ['test', '-coverprofile=' + this.coverageFile]
```
joefitzgerald: Coverage is configurable. We should check to see if coverage is enabled? `atom.config.get('go-plus.test.runTestsWithCoverage')`

lib/build/builder.js
zmb3: Not in this case, we do an extra `i++` below.

lib/build/builder.js
zmb3: Not necessary here, `argparser()` handles white space for us.

lib/build/builder.js
zmb3: See previous comment.

lib/test/tester.js
zmb3: Nothing changed as part of this PR.  The coverage toggle only affects whether we include `-cover` in the `go test` command (for running tests), NOT while running `go test -c` to compile tests (for build errors).

Issue #832, opened 2018-11-14
Golang environment in Docker


dhaavi:
### Prerequisites

* [x] Have you tried launching `atom .` from the terminal in your project's directory?
* [x] Have you verified the output from `go env` is correct? If it is, please include the output in this issue.
* [x] Have you updated Atom to the latest version?
* [ ] Have you tried using [Atom Beta](https://atom.io/beta), which can be run side-by-side with Atom Stable? Is the behavior the same, or different?
* [x] Have you updated your Atom packages to the latest versions?
* [x] Have you read the [FAQ](https://github.com/joefitzgerald/go-plus/wiki/FAQ)?
* [x] Have you searched [the issues](https://github.com/joefitzgerald/go-plus/issues?utf8=%E2%9C%93&q=is%3Aissue) to see if others are experiencing the same issue?

### Description

I am trying to run the complete golang environment in docker, while integrating atom/go-plus into it. 

Host: MacOS Mojave
Container: ubuntu:latest

I imagine this to work as follows:
- go-plus searches for the go binary or other go tools in `PATH`.
- go-plus finds what it is looking for and executes the tool.
- But the tool is just a wrapper script that runs the command in an already running docker container.
- The docker container is completely set up with the go environment and everything works fine in the container.
- The project paths outside/inside container are the same and the network is shared with the host for maximum compatibility (for the beginning).

__The problem:__

go-plus does not pick up the wrapper scripts in the `PATH`, it gives me the "Missing Go Tool" error.

I added the directory with the wrapper scripts to the atom environment by adding
```
process.env.PATH = [process.env.PATH, '/Users/[REDACTED]/fakego/bin'].join(':')
```
to `init.coffee`.

The directory contains wrapper scripts for `delve`, `go`, `gocode`, `godef`, `gofmt`, `gogetdoc`, `goimports`, `golint`, `gometalinter`, `gorename`, `goreturns` and `guru`.

I also tried to set `GOPATH` and/or `GOROOT` there, as well as in the package config interface, to no avail.

### Output from `atom -v && apm -v`

On host:
```
Atom    : 1.32.2
Electron: 2.0.9
Chrome  : 61.0.3163.100
Node    : 8.9.3
apm  2.1.2
npm  6.2.0
node 8.9.3 x64
atom 1.32.2
python 2.7.10
git 2.17.2
```

### Output From `go env`

Ran on host, proxied into container:
```
GOARCH="amd64"
GOBIN=""
[...]

PR #835, opened 2018-11-23
Use coverageFile specified by -coverprofile, if given

dhaavi:
Fixes an issue related to https://github.com/joefitzgerald/go-plus/issues/832:

If a -coverprofile is specified in "Additional test args", then use the given filepath.
Possible improvement: Add a random directory between given dir and file. A static coverage file _may_ cause issues, such as race conditions.

I am not an experienced JS developer, so treat this as a Proof of Concept ;)

### DIFF
--- a/lib/test/tester.js
+++ b/lib/test/tester.js
@@ -276,11 +276,18 @@ class Tester {
   }
 
   createCoverageFile() {
-    this.removeTempDir()
-    if (!this.tempDir) {
-      this.tempDir = fs.realpathSync(temp.mkdirSync())
+    const additionalTestArgs = atom.config.get('go-plus.config.additionalTestArgs')
+    const res = additionalTestArgs.match(/-coverprofile=([A-Za-z0-9/\._-]+)/);
+
+    if (res != null) {
+      this.coverageFile = res[1]
+    } else {
+      this.removeTempDir()
+      if (!this.tempDir) {
+        this.tempDir = fs.realpathSync(temp.mkdirSync())
+      }
+      this.coverageFile = path.join(this.tempDir, 'coverage.out')
     }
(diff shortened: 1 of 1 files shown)


lib/test/tester.js
```diff
@@ -259,26 +259,31 @@ class Tester {
   }
 
   createCoverageFile() {
-    this.removeTempDir()
-    if (!this.tempDir) {
-      this.tempDir = fs.realpathSync(temp.mkdirSync())
+    const additionalTestArgs = atom.config.get('go-plus.config.additionalTestArgs')
+    const res = additionalTestArgs.match(/-coverprofile=([A-Za-z0-9/\._-]+)/);
```
lloiser: This regex won't match windows paths because they use `\` instead of `/`.

lloiser:
Hey @dhaavi, thx for the PR!
Can you please rebase it on master?
Regarding the random dir: IMO that's the user's responsibility because she/he provides this argument and should know what she/he does.

lib/test/tester.js
dhaavi: Changed to `[^ ]+`.

dhaavi: Thas was a bad idea. It's now `[^\s]`.

dhaavi:
Rebased on master and fixed+tested the new regex.

Regarding the random dir: Yes, you are right - it was just an idea I had.

dhaavi:
Is there a timeline for when this PR will be merged?

zmb3:
Sorry, looks like there’s just some linter errors and I’m happy to merge when those are fixed.

dhaavi:
Thanks @joefitzgerald for fixing this for me, just wanted to finally finish this ;)

PR #847, opened 2019-01-06
support revive and golangci-lint linters

### LINKED ISSUE #819: Support of golangci-lint
Hi!
Thank you for the great package. Please add support of [golangci-lint](https://github.com/golangci/golangci-lint). It's the tool similar to gometalinter.

To speed up development, you can take a look at how [existing integrations](https://github.com/golangci/golangci-lint#editor-integration) for another IDEs work.

In short, we just need to be able to run `golangci-lint run` and add flags (e.g. `--fast`) to it.

lloiser:
This PR introduces support for support [`revive`](https://github.com/mgechev/revive)and [`golangci-lint`](https://github.com/golangci/golangci-lint) linters in addition to [`gometalinter`](https://github.com/alecthomas/gometalinter)

closes #819

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -78,7 +78,7 @@ export default {
     this.information = null
     this.implements = null
     this.importer = null
-    this.linter = null
+    this.golinter = null
     this.orchestrator = null
     this.outputManager = null
     this.panelManager = null
@@ -123,7 +123,7 @@ export default {
     )
     this.subscriptions.add(
       this.orchestrator.register('linter', (editor, path) => {
-        if (this.linter) this.linter.lint(editor, path)
+        if (this.golinter) this.golinter.lint(editor, path)
         return
       })
     )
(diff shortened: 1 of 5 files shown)


lib/main.js
```diff
@@ -426,8 +426,8 @@ export default {
   consumeLinter(registry: any) {
     this.buildLinter = registry({ name: 'go build' })
     this.subscriptions.add(this.buildLinter)
-    this.gometalinterLinter = registry({ name: 'gometalinter' })
-    this.subscriptions.add(this.gometalinterLinter)
+    this.linterLinter = registry({ name: 'go linter' })
```
zmb3: s/linterLinter/linter/g

package.json
```diff
@@ -397,9 +397,21 @@
...
+            "revive",
+            "golangci-lint"
+          ],
+          "order": 1
+        },
         "args": {
-          "title": "gometalinter Arguments",
-          "description": "Arguments to be passed when invoking `gometalinter`. Please ensure the `--json` argument is always included. Arguments are comma-separated.",
+          "title": "Linter Arguments",
+          "description": "Arguments to be passed when invoking the provided linter. Arguments are comma-separated.",
```
zmb3: For `additionalBuildArgs` and `additionalTestArgs` we use a config type of `string` and parse them with `yargs-parser`. 

We should probably be consistent for all config properties that represent arguments to feed into an external tool.

I prefer `string` over `array` as the string is more natural and looks exactly as you'd type it into a shell, rather than comma separating the args just to appease atom's config parser.

lib/main.js
lloiser: `this.linter` already exists, which is the go-plus `Linter` instance. Maybe I should rename `linterLinter` to `linterProvider` or something else...

package.json
lloiser: Makes sense. The only thing is backward compatibility. I would have to use `args.join(' ')` in case the args are still an array in the config. Maybe even write them back to the config afterwards.

PR #851, opened 2019-01-13
Fix builds

lloiser:


zmb3:
What one of the commits was the fix?  I spent a ton of time trying to figure this out and gave up.

lloiser:
First I tried to use the exact same macos image that the last green build had used (df63c95b2f90b7016b650164721c3157b96b112e) - no success
Then I added a whole lot of logs using jasmine assertions (5cb7b27950e57bd43ae2e9ecc82afb005c0bf11d) because console.log did not work - [eventually this got me into the right direction](https://travis-ci.org/joefitzgerald/go-plus/jobs/478979882#L175)

Turns out that `fsevents` was incompatible on macos for whatever reason (I do not have a mac at hand so I could not try...)
Luckily `fsevents` is only a dependency of [`babel-plugin-add-module-exports`](https://github.com/59naga/babel-plugin-add-module-exports).
This babel plugin transforms default exports into something that is easier to 
consume within nodejs.

The fix seemed rather easy: _just_ get rid of `export default` and `fsevents` will also vanish.
Unfortunately this was tricker than I initially thought because several files did not use flowtype at all and the main file did not type check properly.
So I ended up flowtyping everything, getting green tests and fixing some things on the way 😁 

My next step is to clean up all these intermediate commits by rebasing/squashing everything into a few commits only. No one needs those in the git history, right?

lib/format/formatter.js
```diff
@@ -50,7 +50,7 @@ class Formatter {
       console.log('skipping format, could not find tool', tool) // eslint-disable-line no-console
       return null
     }
-    const options = this.goconfig.executor.getOptions('project')
+    const options = this.goconfig.executor.getOptions('project', editor)
```
lloiser: shouldn't the formatter run in the context of the file it is formatting? So use `file` instead of `project`?

zmb3: If format tool is goimports and the project is using modules then the working dir needs to be project dir in order to resolve module dependencies and suggest the correct imports.

lib/panel/panel-manager.js
```diff
@@ -194,13 +180,15 @@ class PanelManager {
       return
     }
 
-    this.goPlusStatusBar = new GoPlusStatusBar({
+    const goPlusStatusBar = new GoPlusStatusBar({
```
lloiser: this code path is only called from main.js but the function there is not called. Should I remove it or actually use it, what do you think?

lloiser: Afaik this was only used to display some kind of busy indicator. It is not updated anywhere. I will remove it :)

zmb3: I think this was used to display the signature of the function under the cursor in the status bar.

This was powered by gocode, and I'm not sure if mdempsky/gocode even supports it anymore.  Plus, now that we have hover tooltips the future is less useful.   I'm all for removing it.

.travis.yml
```diff
@@ -49,10 +49,10 @@ git:
 addons:
   apt:
     packages:
-    - build-essential
```
zmb3: Did this change actually do anything? Pretty sure it shouldn't be necessary.

lib/get/get-manager.js
```diff
@@ -225,7 +225,7 @@ class GetManager {
       return { success: false, result: null, results: null }
     }
 
-    const executorOptions = this.goconfig.executor.getOptions('file')
```
zmb3: Why the change from file to project here?

lloiser: It did not make any sense to me to run `go get` in the context of the currently open file. This could have been any file without understanding it, and even falls back to the 'project' options in case no file is currently open.
I can revert it if necessary.

zmb3: no need to revert, was just curious

zmb3:
Looks good to merge if you're finished with it.

PR #855, opened 2019-01-17
only run `go install/build` or `go test`

lloiser:
On smaller projects, running both may seem like a reasonable approach because both commands finish in less than 1 second. For larger projects, though a simple go build can take several seconds especially if you include C code.

TODOs:

- [x] The only downside I can think of at the moment is autocompletion... I will have to test this for a while. Update: go test supports installing all dependencies via -i
- [x] Maybe kill previous install/test of the same folder if it has not finished yet.

### DIFF
--- a/lib/build/builder.js
+++ b/lib/build/builder.js
@@ -73,6 +73,8 @@ class Builder {
     }
     this.deleteMessages()
     const options = this.goconfig.executor.getOptions('file', editor)
+    options.timeout =
+      (atom.config.get('go-plus.config.buildTimeout'): any) || 10000
     const cmd = await this.goconfig.locator.findTool('go')
     if (!cmd) {
       throw new Error('cannot find go tool')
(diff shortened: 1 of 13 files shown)


zmb3:
This one good to go too?

lloiser:
I have tested it for a while now so it seems to work :)

lib/build/builder.js
```diff
@@ -239,14 +233,14 @@ class Builder {
           // we'll take care of this one, skip over the -o flag
           i++
           continue
-        } else if (parsed[i] === '-c') {
+        } else if (parsed[i] === '-c' || parsed[i] === '-i') {
```
zmb3: Which version of Go added the `-i` flag to `go test`?  Is it supported in 1.10?

lloiser: [1.9](https://github.com/golang/go/blob/release-branch.go1.9/src/cmd/go/internal/test/test.go#L104), [1.10](https://github.com/golang/go/blob/release-branch.go1.10/src/cmd/go/internal/test/test.go#L157) and obviously [1.11](https://github.com/golang/go/blob/release-branch.go1.11/src/cmd/go/internal/test/test.go#L154).

Interestingly this even dates back to [1.0](https://github.com/golang/go/blob/release-branch.go1/src/cmd/go/test.go#L62)!

PR #860, opened 2019-01-20
beautify datatips

lloiser:
Before:
image

After:
image


zmb3:
Looks like we just have to update the assertion in the specs and this will be good to merge.

PR #889, opened 2019-05-22
Add automatic retry feature to get-manager

niruix:
This pull request will add automatic retry feature to `get-manager` in order to improve the robustness of updating in an unstable network.

It will check the exitcode of `await this.goconfig.executor.exec`, namely `ExecResult.exitcode`, to determine whether or not the `go get` command is successful (see below). If not, it will go a loop and retry that command.

    if (!r.exitcode) {
        getOutput += '$ go ' + argsStr + os.EOL
        break
    }

User will be notified through the output text, for example, when retrying, the status line will be updated to:

    Retrying(3/6): $ go get -u github.com/derekparker/delve/cmd/dlv

Format: `Retrying(<How many time already tried>/<Total retry limit>)`

If later the command has failed and given up, the line will be updated to:

    Failed: $ go get -u github.com/derekparker/delve/cmd/dlv
    <Error info if available>

An Atom package option has also been added to allow user to configure the retry limit (Retry how many times). The default value of the option has been set to only 1 as retry does not resolve all the problem (It can only fix temporary problems, but temporary problems is annoying and often).

The option also been programmatically ensured cannot less than 1 using code:

    if (!retries || typeof retries !== 'number' || retries < 1) {
        retries = 1
    }

### DIFF
--- a/lib/get/get-manager.js
+++ b/lib/get/get-manager.js
@@ -231,8 +231,8 @@ class GetManager {
       executorOptions.timeout = timeout
     }
 
-    let retries = atom.config.get('go-plus.get.retries')
-    if (!retries || typeof retries !== 'number' || retries < 1) {
+    let retries = Number(atom.config.get('go-plus.get.retries'))
+    if (retries < 1) {
       retries = 1
     }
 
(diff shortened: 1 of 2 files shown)


reinit:
Sorry, I've failed to please both CIs.

nirui:
Discarded: No longer use Atom.

PR #906, opened 2019-06-04
WIP: support Go language server (gopls)

joefitzgerald:
WIP

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,6 @@ install:
   - go get -u golang.org/x/tools/cmd/goimports
   - go get -u golang.org/x/tools/cmd/gorename
   - go get -u github.com/sqs/goreturns
-  - go get -u github.com/mdempsky/gocode
   - go get -u github.com/alecthomas/gometalinter
   - go get -u github.com/zmb3/gogetdoc
   - go get -u github.com/zmb3/goaddimport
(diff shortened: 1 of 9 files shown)


tekkamanendless:
Oh yes!  Is there any help that you guys need with getting `gopls` rolled in?  Ever since Go modules, dealing with autocompletion has been a nightmare.  I'm pumped.

zmb3:
Not actively working this, but if somebody wants to pick it up feel free.

The auto-languageclient package gets you almost all of the way there.  The challenge you'll run into is that atom-ide-ui is no longer maintained, so if the UI doesn't support something, you've got to try to resurrect a dead project or start over.

Alternatively, you could try https://github.com/MordFustang21/ide-gopls.

tekkamanendless:
I would think (perhaps naively) that it should just be a replacement operation (replace the call to `gocode` with the equivalent call to `gopls`).  Since we're not trying to add anything new or fancy, tweaking `atom-ide-ui` shouldn't be necessary, right?

Do you have any notes on what doesn't work with this change so far?  I'll try to pull it down when I get some down time on this project.

zmb3:
You want to actually use the auto language client package, as opposed to replacing the gocode call only. This will register to provide more than just autocomplete from gopls (lint rules, go to definition, etc)

PR #951, opened 2019-10-07
Configurable Guru Timeout

robbawebba:
This commit adds a user-configurable timeout value to all guru commands.

closes #951 

Two concerns I would like to address about the proposed changes:
* I am  encountering an issue with guru when running tests locally (both on my branch and on the HEAD of master): `guru: no initial packages were loaded`. I'm unsure if this is due to my local dev environment, so I wanted to see if the tests pass in the CI environment.
* It is proving to be very difficult to test the highlight provider guru timeout. this guru operation seems to work pretty quickly in the test environment. When I set a timeout of 1ms in the package settings during normal usage, the command does not seem to finish within 1ms and no identifiers in the editor are highlighted. when I set a timeout of 1ms in a spec, the highlight provider's highlight function seems to finish before the timeout, and valid highlight ranges are returned by the function. I'm not sure if it's going to be possible to test the timeout of the highlight provider.

### DIFF
--- a/lib/highlight/highlight-provider.js
+++ b/lib/highlight/highlight-provider.js
@@ -42,7 +42,10 @@ class HighlightProvider {
     if (!args) return null
 
     const options = {}
-    options.timeout = 30000
+    const t = atom.config.get('go-plus.guru.timeout')
+    const timeout = typeof t === 'number' ? t : 30000
+    options.timeout = timeout
+
     const archive = buildGuruArchive(editor)
     if (archive && archive.length) {
       options.input = archive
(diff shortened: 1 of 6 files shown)


zmb3:
Looks like a couple tests are broken, but otherwise LGTM.

robbawebba:
Hi @zmb3, thanks for taking a look!
I'm running into issued with running the specs locally, which makes the issues difficult to debug. I'm running several issues with guru locally that do not seem to be a problem in the CI environment.
For example, most guru commands that are run through the specs fail with this error:
```
guru: no initial packages were loaded
```

Have you seen this before while running tests locally?

zmb3:
I haven't seen that locally, though I admit it's been a while since I've tried.  That sounds to me like a gopath or working directory problem.

